PR #4, opened 2017-12-08
Safety control

scholvat:


Autopilot/Inc/main.h
```diff
@@ -1,6 +1,6 @@
 /**
   ******************************************************************************
-  * File Name          : main.h
+  * File Name          : main.hpp
```
edenfrosst: File name is not, in fact, `main.hpp`

Autopilot/Inc/main.h
```diff
@@ -48,6 +48,8 @@
 /* Define to prevent recursive inclusion -------------------------------------*/
 #ifndef __MAIN_H
 #define __MAIN_H
+  /* Includes ------------------------------------------------------------------*/
+
```
edenfrosst: Useless cube changes

Safety/Src/PWM.c
```diff
@@ -99,9 +105,10 @@ void HAL_TIM_IC_CaptureCallback(TIM_HandleTypeDef *htim) {
         uint16_t time_diff = HAL_TIM_ReadCapturedValue(htim, TIM_CHANNEL_1);
 
         if (ppm_index != 0) { // the first edge doesn't give us any data
-            capture_value[ppm_index - 1] = time_diff - PPM_OFFSET;
-        }
-        ppm_index = (ppm_index + 1) % (PPM_NUM_CHANNELS + 1); // index should reset, but just in case
+            capture_value[ppm_index - 1] = (time_diff - PPM_OFFSET) / 2.5; //Divide by 5 to keep PPM and PWM at same scale
```
edenfrosst, line 108: Why do they need to be at the same scale?

Safety/Src/main.c
```diff
@@ -108,26 +107,23 @@ int main(void)
...
+
+  Safety_Init();
 
   /* USER CODE END 2 */
 
   /* Infinite loop */
   /* USER CODE BEGIN WHILE */
   while (1)
   {
+    Safety_Run();
```
edenfrosst: `while (1)` inside a `while (1)`?

Safety/Src/tim.c
```diff
@@ -444,7 +444,7 @@ void HAL_TIM_Base_MspInit(TIM_HandleTypeDef* tim_baseHandle)
     HAL_GPIO_Init(PPM_GPIO_Port, &GPIO_InitStruct);
 
     /* TIM14 interrupt Init */
-    HAL_NVIC_SetPriority(TIM14_IRQn, 0, 0);
+    HAL_NVIC_SetPriority(TIM14_IRQn, 1, 0);
```
edenfrosst, line 447: Does lowering the priority of the TIM IRQ make it possible to lose PPM data during a SPI IRQ?

Safety/Src/safety_control.c
```diff
@@ -0,0 +1,56 @@
...
+                    if(PPM[i]>PWM_MAX) PPM[i] = PWM_MAX;
+                }
+                offset = offset << 1;
+            }
+        }
+        PWM_SetAll(PPM);
+    }
+}
+
+uint16_t Safety_isManual(int16_t PWM_Value){
```
edenfrosst: Why is this a uint16? Should be of type `bool`. Include `<stdbool.h>` if it's not already.

Safety/Src/main.c
```diff
@@ -108,26 +107,23 @@ int main(void)
...
 
   /* USER CODE BEGIN 2 */
-  PWM_Init();
 
   debug("\r\n\r\nStarting up...");
   debug("Compiled on %s at %s", __DATE__, __TIME__);
 
   HAL_GPIO_WritePin(LED1_GPIO_Port, LED1_Pin, GPIO_PIN_RESET);
   HAL_GPIO_WritePin(LED2_GPIO_Port, LED2_Pin, GPIO_PIN_SET);
   HAL_GPIO_WritePin(LED3_GPIO_Port, LED3_Pin, GPIO_PIN_SET);
```
edenfrosst, line 116: Need to use the LEDs for something useful.

Safety/Src/safety_control.c
```diff
@@ -0,0 +1,56 @@
...
+            dataTX->PWM[i] = PPM[i];
+        }
+        dataTX->safety_level = isManual;
+
+        //if not manual, change values from input PPM to Autopilot PWM values
+        if(!isManual){
+            uint8_t i;
+            uint16_t offset =1;
+            for(i=0; i<PPM_NUM_CHANNELS;i++){
+                if( (dataRX->autonomous_level & offset) !=0){
```
edenfrosst: `!=0` is redundant

edenfrosst (changes requested):
Some things I'm not sure about. Some of the formatting is broken, too. If you could run `clang-format` on the non-Cube files you changed, that would be great. There's a config file for it in the root of the repo.

Safety/Src/PWM.c
scholvat: It makes a lot more sense for them to be at the same scale. That means I can just call PWM_SetAll(PPM_Get()) and it works. You lose some accuracy, but that accuracy is irrelevant anyways since it is mostly being used for outputting PWM

Safety/Src/safety_control.c
scholvat: It is due to the size needed for the SPI struct... Prevents me from having to cast it from bool to uint16 later on in the code, but could be argued either way. Your call.

Safety/Src/tim.c
scholvat: Its possible but unlikely, and SPI doesn't work at all if you keep them the same priority.

Safety/Src/main.c
scholvat: Out of scope for this PR, but yes we didn't spend all that money on LEDs for nothing :P

scholvat: I put an asana task for it, a good first year task probably

scholvat:
Add clang_format stuff and Cmake stuff to README @ianjfrosst

scholvat:
@ianjfrosst files cleaned up

Safety/Src/safety_control.c
edenfrosst: Cast should happen, either explicitly or implicitly, when inserting data into the structure. You're already doing implicit casts from bool (&&) to uint16_t.

Safety/Src/tim.c
edenfrosst: Right. Just don't want to get in a state where the Autopilot can lock up the safety controller via SPI. Should spend some time checking this (ex. by injecting faulty SPI protocol).

Safety/Drivers/Peripherals/Src/Interchip_S.c
```diff
@@ -6,19 +6,30 @@
...
+void HAL_SPI_TxRxCpltCallback(SPI_HandleTypeDef *hspi){
   //Once packet recieved, listen for next packet
   if(hspi->Instance == SPI1){
     HAL_SPI_TransmitReceive_IT(&hspi1,(uint8_t *)dataTX,(uint8_t *)dataRX, sizeof(Interchip_AtoS_Packet)/sizeof(uint16_t));
   }
+}
 
+void HAL_SPI_ErrorCallback(SPI_HandleTypeDef *hspi){
+  if(hspi->Instance == SPI1){
+    HAL_SPI_TransmitReceive_IT(&hspi1,(uint8_t *)dataTX,(uint8_t *)dataRX, sizeof(Interchip_AtoS_Packet)/sizeof(uint8_t));
```
edenfrosst: Why is this one `sizeof(uint8_t)`?

scholvat: because i fucked up

edenfrosst (approved):

PR #5, opened 2017-12-15
Clean up Autopilot file structure

edenfrosst:


scholvat (commented):
It seems that this file structure breaks stmcube. How are you planning to use Cube after changing the file structure?

Also if you could include a little blurb in the README about the file structure that would be helpful

edenfrosst:
Cube generated code has some issues. Mainly formatting & structure, makes it unfriendly to development. Plan was to use Cube as a "jumping-off" point, to give us a place to start. All this does that breaks Cube is merge Inc/ into Src/, and device-specific files into Dev/. Were we planning on using Cube for much more?

scholvat:
I think continuing to use cube would be a good idea. Its a good way to easily change peripheral settings, and it will be very helpful for doing new hardware revisions of the board since we should mainly just need to change the pin outs in cube. Why don't we just keep /Inc and /Src separate so that cube will still work? I don't see much of a benefit of merging them into 1 directory.

PR #6, opened 2017-12-19
Added Safety Watchdog

scholvat:


Safety/Src/safety_control.c
```diff
@@ -16,10 +17,12 @@ void Safety_Init() {
   Interchip_Init(dataTX, dataRX);
 }
 
-void Safety_Run() {
-  while (1) {
-    volatile int16_t* PPM_input = PPM_Get();
-    for (uint8_t i = 0; i < PPM_NUM_CHANNELS; i++) {
+void Safety_Run(){
```
edenfrosst: Should have signature `void Safety_Run(void)`, same in the header

edenfrosst (approved):
All good. Just needs fix+formatting.

Safety/Src/safety_control.c
scholvat: why is that needed?

PR #7, opened 2017-12-19
Put Interchip on new thread

scholvat:


Autopilot/Drivers/Peripherals/Src/Interchip_A.c
```diff
@@ -1,21 +1,31 @@
 #include "Interchip_A.h"
 #include "debug.h"
 #include "spi.h"
+#include "freertos.h"
```
edenfrosst: `#include "cmsis_os.h"` is what you're looking for.

Autopilot/Src/freertos.c
```diff
@@ -118,6 +124,18 @@ void StartDefaultTask(void const * argument)
   /* USER CODE END StartDefaultTask */
 }
 
+/* Interchip_Run function */
+__weak void Interchip_Run(void const * argument)
```
edenfrosst: This is definitely not the correct way to do this. **Declare** the function in `Interchip_A.h` (note that it must take a `void const *` parameter than can be casted any way you like), then call it from here. **OR** `#include "Interchip_A.h` at the top of this file and write the loop here. You shouldn't be writing `__weak` functions outside of a library context.

scholvat: I used the cube autogeneration for this, I can define it as weak, external, or normal. I'll fix the function parameters (I'm confused why it works without the correct signature, but it does). I'll change it to not use weak if that isn't conventional.

scholvat:
@ianjfrosst changed to extern, take a look

Autopilot/Drivers/Peripherals/Inc/Interchip_A.h
```diff
@@ -1,5 +1,7 @@
 #include "stm32f7xx_hal.h"
 #include "Interchip.h"
 
-void interchipInit(Interchip_AtoS_Packet*, Interchip_StoA_Packet*);
-HAL_StatusTypeDef Interchip_Update();
\ No newline at end of file
+#define INTERCHIP_TRANSMIT_DELAY 5
+
+Interchip_StoA_Packet Interchip_GetData();
```
edenfrosst: Needs `void` signature

Autopilot/Src/wwdg.c
```diff
@@ -84,7 +84,7 @@ void HAL_WWDG_MspInit(WWDG_HandleTypeDef* wwdgHandle)
     __HAL_RCC_WWDG_CLK_ENABLE();
 
     /* WWDG interrupt Init */
-    HAL_NVIC_SetPriority(WWDG_IRQn, 5, 0);
+    HAL_NVIC_SetPriority(WWDG_IRQn, 4, 0);
```
edenfrosst: This maybe should be priority 0. 5 is set as the default for ISRs that might interact with threads and other FreeRTOS functions.

Autopilot/Drivers/Peripherals/Src/Interchip_A.c
```diff
@@ -1,21 +1,50 @@
...
+
+int16_t *Interchip_GetPWM(void) { return dataRX->PWM; }
+void Interchip_SetPWM(int16_t *data) {
+  osMutexWait(Interchip_MutexHandle, 0);
+  for (uint8_t i = 0; i < PWM_NUM_CHANNELS; i++) {
+    dataTX->PWM[i] = data[i];
+  }
+  osMutexRelease(Interchip_MutexHandle);
+}
+uint16_t Interchip_GetSafetyLevel(void) { return dataRX->safety_level; }
```
edenfrosst: Should have 1-2 newlines between functions

Common/Inc/Interchip.h
```diff
@@ -2,15 +2,16 @@
 * Interchip packets
 * Packets should be a multiple of 16 bits
 */
+#include "Common.h"
 
 typedef struct
 {
-  int16_t PWM[12];
+  int16_t PWM[PWM_NUM_CHANNELS];
```
edenfrosst, line 9: No guarantee that we have the same number of PPM inputs as we do outputs (. However, this does make the packets the same size. No harm in making the last 4 data 0.

Autopilot/Src/freertos.c
```diff
@@ -113,11 +126,22 @@ void StartDefaultTask(void const * argument)
...
+{
+  /* USER CODE BEGIN Attitude_Run */
+  /* Infinite loop */
+  for(;;)
+  {
+    osDelay(1);
+  }
+  /* USER CODE END Attitude_Run */
+}
+
```
edenfrosst: Attitude code isn't relevant to this PR

edenfrosst (approved):

PR #10, opened 2018-01-23
Add basic Linux env setup scripts for Autopilot subproject

stephen-zhao:
See commit history (and long descriptions) for explanation of what these are for.

Basically, if you have all your arm-none-eabi-gdb binaries installed (or make a symlink here otherwise) at "~/opt/arm-none-eabi-gdb/current/bin/", then running the setenv_autopilot.sh script should set up a bunch of useful aliases, especially the ones used to flash and debug the target. It is suggested to create a personal alias, like "ezpap", for the setenv_autopilot.sh script.


Autopilot/stm32f7-openocd.cfg
```diff
@@ -0,0 +1,12 @@
+# This is an STM32F7 discovery board with a single STM32F756NGH6 chip.
+# http://www.st.com/web/catalog/tools/FM116/SC959/SS1532/LN1848/PF261641
+
+# This is for using the onboard STLINK/V2
+source [find interface/stlink-v2.cfg]
+
+transport select hla_swd
+
+# increase working area to 256KB
+set WORKAREASIZE 0x40000
```
edenfrosst, line 10: What does `WORKAREASIZE` represent? Is it the flash space of the chip? Might need to be different for our board.

stephen-zhao: hmm.. From [here](http://openocd.org/doc/html/Config-File-Guidelines.html):

"Work areas are small RAM areas associated with CPU targets. They are used by OpenOCD to speed up downloads, and to download small snippets of code to program flash chips."

stephen-zhao: This cfg file (and so the 0x40000) was copied from a config file for the stm32f7discovery, which uses the same CPU I believe.

Autopilot/autopilot.envrc
```diff
@@ -0,0 +1,59 @@
...
+    echo "Press any key to abort environment change..."
+    read input
+    exit 1
+fi
+
+# Firstly, source bashrc
+[[ -e ~/.bashrc ]] && source ~/.bashrc
+
+# Add ARM compilers to PATH
+export PATH=$PATH:$HOME/opt/gcc-arm-none-eabi/current/bin
```
sergei1152, line 20: This isn't a standard linux folder (~/opt). However /opt is, although it requires root privelages to access. Why did you select ~/opt? Does the toolchain get installed there by default on Mac/Linux?

stephen-zhao: [This](https://gnu-mcu-eclipse.github.io/toolchain/arm/install/) particular website (last section) says to install there. The site is linked from the [Atlassian UWARG docs site](https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/25526274/Getting+Started) where I followed it to install there. We can change it, but then we'd need our own set of instructions for installation (or I can write another script to automatically download and install it, openocd et al, to a better location).

Autopilot/autopilot.envrc
```diff
@@ -0,0 +1,59 @@
...
+alias gdbflash="arm-none-eabi-gdb \
+     -se \"$AUTOPILOT_ROOT/build/Autopilot.elf\" \
+     -x \"$AUTOPILOT_FLASH_GDB_CMDS\""
+
+# Set up navigational aliases
+alias cdzpap="cd $AUTOPILOT_ROOT"
+alias cdzpapsrc="cd $AUTOPILOT_ROOT/Src"
+
+# Set up build aliases
+alias buildzpap="cd $AUTOPILOT_ROOT && make"
```
edenfrosst: Possibly should use CMake instead of the old Makefile?

Autopilot/Tools/build.sh
```diff
@@ -0,0 +1,28 @@
...
+if [[ $# > 0 ]]; then
+    echo "with cmake parameters: $@"
+fi
+echo ""
+
+# Build commands
+mkdir -p build || die $?
+cd build || die $?
+cmake ../ -DCMAKE_TOOLCHAIN_FILE=STM32F765xG.cmake -G "Unix Makefiles" $@ || die $?
+make || die $?
```
edenfrosst, line 23: Instead of doing `|| die` everywhere, why not just use `set -eu`?

stephen-zhao: The custom fatal error handling, like displaying a message is nice to have. We don't have that many build steps at the moment so it should be fine.

stephen-zhao:
On another note, I will submit another PR later to generalize these build scripts to also work with the Safetly subproject.

edenfrosst (approved):

PR #14, opened 2018-06-20
Update Autopilot configs for full C++ support

edenfrosst:
This PR also includes some updates to the `build.bash` script for easier compile options.


scholvat:
Can you update the README with instructions for compiling using the bash script?

edenfrosst:
It's got a usage string....

scholvat:
There needs to be instructions in the README on what commands to run to build and program the project, regardless of how trivial it might be. Also include instructions on how to run bash scripts on Windows, since it isn't configured by default in windows

scholvat:
What is up with the test option? It seems to be in the bash script but I don't see anything in the Cmake to configure testing, and its throwing an error when I try to run build.bash -t (error: No rule to make target 'test')

edenfrosst:
Most of the options from that script are pulled from the one in `computer-vision`. I've been working on getting basic linting running for the project, as well as a test framework for (semi-)automated unit and integration tests. That stuff isn't ready yet, so I haven't included it here.

edenfrosst:
I might add a (simpler) PowerShell or Batch script to run on Windows.

scholvat:
Still, some documentation is needed for how to compile and flash. The actual script and clang formatting looks fine (with the testing features coming in the future), but its pretty pointless to have those scripts if there is no documentation on how to use them or how to use the clang formatter.

edenfrosst:
It's a work in progress. If you want the `main.cpp` functionality (which is what you asked for), it's here. The rest of the tooling isn't finished yet.

scholvat (approved):
I'll go ahead and merge it in now then, the code seems fine and I verified it does program the board. I'm going to add the documentation on docs.uwarg

PR #28, opened 2020-05-21
Integrated unit test framework and added travis-Ci support

Fberbari:
With this commit, we have the ability to write unit tests for the
Autopilot and run them via a command line option. Along with this, the
travis-ci config file has been updated to instruct travis to run all
unit tests on windows, linux, and macos. It has also been updated to
attempt to build the code using the embedded compiler on macos and
windows (previously travis only made sure things built on linux).

Quick breakdown of why each file was added, deleted, or changed:

The gitignore file now ignores the new testBuild directories that are
created when unit testing is taking place.

The travis config file is updated to implement the features mentioned in the
first paragraph.

The CmakeLists file is updated to add the ability to perform a unit test build.

The build.bash file is updated to allow the new -t option, which instructs
cmake to build the unit tests, rather than build for the hardware.

The 2 Test_*_Skeleton files added to the UnitTestSrc directory are meant to
serve as examples for writing unit tests and also allow things to compile during the
unit test build (since the Altimeter and Imu modules already had partial
implementations that attempted to talk to hardware, they needed to be stubbed out).

The testMain.cpp file serves as a main function for the executable produced
when running unit tests. It should not be messed with.

The fff.h file is the complete fake functions framework, which provides us with
the ability to mock C stuff (while googletest's gmock gives us the ability to mock C++
stuff). It is being linked in to the unit test builds.

Lastly, the /Safety/Common folder is deleted. It was never meant to exist. It
was a symbolic link created by Cmake to cope with a bug that has already been
fixed. Since the build system is being updated in this commit, it is a good
opportunity to get rid of it.


heyitslu (approved):
Ran autopilot build.bash script on Windows through Git Bash. -t for unit testing works as long as MinGW is installed using chocolatey and googletest root folder has been added as environment variable GTEST_ROOT.

avlepage:
Ran build.bash from cl on MacOS with gnu-arm-gcc. -t works without a hitch, tests pass thus far. My license for CLion has expired, otherwise I would have run it through CLion to double check (will look into this).

If I can wrangle another free license from jetbrains then I'll check it. If not, then it really won't matter all that much. New kiddies can learn to use the command line; it's good for them. Also, I can't imagine the IDE compile will go differently, but you never know.

TLDR: pass, works great on command line! Nonsequitur: we may need to axe CLion if they only do one year free for students.

avlepage (approved):
Turns out I didn't know how to properly review. TLDR: Pass, see the prev. comment for the long and short.

PR #29, opened 2020-05-27
Added IMU header file, along with an IMU Mock file.

heyitslu:
I've added 3 new files:
IMU.hpp - IMU header with IMU class and IMUData_t struct
IMU_Mock.hpp - Mock IMU interface to be used during unit tests
Test_IMU.cpp - File to contain IMU unit tests

Deleted EMPTYFILE.txt


heyitslu:
Fixed the failed unit test on Linux issue. The old MOCK_METHODn macro (where n is the number fn arguments seemed to be incompatible) from gtest 1.7.0 with Linux.

Replaced 
```
- sudo apt-get install libgtest-dev
- sudo apt-get install google-mock
```
with curl install method to install gtest 1.10.0. (apt-get install is not up to date and pulls in 1.7)

Autopilot/Test/Inc/IMU_Mock.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef IMU_MOCK_HPP
+#define IMU_MOCK_HPP
+
+#include "gmock/gmock.h" // Brings in gMock
```
Fberbari: gmock has been installed as a system library, so should be <gmock/gmock.h>.

.travis.yml
```diff
@@ -20,19 +20,27 @@ matrix:
...
-        - sudo make
-        - sudo cp *.a /usr/lib
-        - cd /usr/src/gmock
-        - sudo cmake CMakeLists.txt
+        # - cd /usr/src/gtest
+        # - sudo cmake CMakeLists.txt
+        # - sudo make
+        # - sudo cp *.a /usr/lib
+        # - cd /usr/src/gmock
+        # - sudo cmake CMakeLists.txt
```
Fberbari: No need to clutter up with the commented lines. We have the commit history if we ever need to revert.

Autopilot/Inc/IMU.hpp
```diff
@@ -0,0 +1,51 @@
+/**
+ * IMU Sensor Functions and Part Number Selection
+ *
```
Fberbari: Wanna add your name as an author? :P

Autopilot/Inc/IMU.hpp
```diff
@@ -0,0 +1,51 @@
+/**
+ * IMU Sensor Functions and Part Number Selection
+ * 
+ */
+
+//#pragma once
```
Fberbari: We should either go with #pragma once or the #ifndef stuff. No need to keep the other commented.

Autopilot/Test/Inc/IMU_Mock.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+class MockIMU : public IMU {
+    public: 
+        MOCK_METHOD(void, Init, ());
+        MOCK_METHOD(void, Begin_Measuring, ());
+        MOCK_METHOD(void, GetResult, (IMUData_t *Data));
+
+};
+
+#endif
```
Fberbari: Everything prior to c++ 11(which happens to be the standard we compile with, and why this doesn't cause warnings), requires a new line at the end of every file. May as well add one, for whatever backwards compatibility and so we don't see this ugly red circle during reviews. (You can get your text editor to add one automatically.)

Autopilot/Test/Src/Test_Imu_skeleton.cpp
```diff
@@ -3,6 +3,11 @@
 #include "i2c.h"
 #include "mpu9255.h"
 
+#include "IMU_Mock.hpp"
+#include "IMU.hpp"
+
```
Fberbari, line 8: Doesent #include <gtest/gtest.h> need to be here ? I guess not since this compiles lol, but seems weird, pretty sure the ::testing namespace comes from those headers anyway. This may be some Cmake magic that's solving the dependencies, in which case it would be best to include gtest, so things dont break if the build system changes at all in the future.

heyitslu: #include <gtest/gtest.h> is the first line of the file

Fberbari (approved):

avlepage (approved):

avlepage:
Looks good to me

PR #31, opened 2020-05-31
Airspeed sensor interface and Mock infrastructure for testing

avlepage:
Created header files for the airspeed sensor and sensor mocker.

Added a test skeleton for the airspeed sensor.


Autopilot/Inc/airspeed.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+        /**GetResult should:
+         *  1. Transfer raw data from variables to struct
+         *  2. Update utcTime and status values in struct as well
+         * 
+         *  Potentially:
+         *  ensure that data acquired makes sense, has been
+         *  gathered recently within reason (past 10s?)
+         * */
+        virtual void GetResult(airspeedData_t *Data); //
```
Fberbari: "//" at the end of the line should be removed

Autopilot/Test/Inc/airspeed_Mock.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+class MockAirspeed : public airspeed {
+	public:
+
+		MOCK_METHOD(void, Init, ());
+		MOCK_METHOD(void, BeginMeasuring, ());
+		MOCK_METHOD(void, GetResult, (airspeedData_t *Data));
+};
+
+#endif
```
Fberbari: There should be a new line at the end of this file. (C++ standards prior to C++ 11 recognize no new line as a problem, as do posix standards).

Autopilot/Inc/airspeed.hpp
```diff
@@ -0,0 +1,59 @@
...
+    Currently there is only one airspeed sensor used, if this is 
+    changed to a different sensor, part numbers and a selection
+    option should be added
+*/
+
+struct airspeedData_t 
+{
+    double airspeed;    // in m/s
+
+    float utcTime;      // 4 Bytes. Time in seconds since 00:00
```
Fberbari: There should be some sort of sensorStatus field (Lucy and I made it an integer) so we can report any error codes.

Also, there should be an isDataNew flag (probably a bool) so a caller can check whether data has been updated since the last time they called this function.

Autopilot/Inc/airspeed.hpp
```diff
@@ -0,0 +1,59 @@
...
+*   per second
+*
+*    Author: Annie LePage
+**/
+
+
+#ifndef AIRSPEED_HPP
+#define AIRSPEED_HPP
+
+#include <stdint.h>
```
Fberbari: I'm not a fan of including libraries that don't need to be there. It doesn't look like you need the stdint library in this file.

Autopilot/Inc/airspeed.hpp
```diff
@@ -0,0 +1,59 @@
...
+};
+
+class airspeed 
+{
+    public:
+        /**
+         *  Initializes ADC in order to recieve sensor readings 
+         *  (pitot tube reading)
+         * */
+        virtual void Init();
```
Fberbari: Had some discussion with Lucy about this. We might aswell make these pure virtual methods, since it's *slightly* safer and more explicit. https://www.youtube.com/watch?v=ndz3EHpFEZc&t=46s is a good explanation of virtual vs pure virtual.

Pure virtual would look like:
virtual void Init() = 0;

Fberbari (commented):
Sweet, a few little things, but looking good!

Autopilot/Inc/airspeed.hpp
avlepage: Sounds good

Autopilot/Inc/airspeed.hpp
avlepage: You're right, I had thought about using integers and then decided against

Autopilot/Inc/airspeed.hpp
avlepage: I can do that

heyitslu (approved):
The PR looks good! Just some minor fixes that Anthony already pointed out :)

Fberbari (approved):
Sweet

PR #34, opened 2020-06-04
Created SensorFusion Interface file + modified file structure to incl…

heyitslu:
…ude a folder for AttitudeManager modules.

### DIFF
--- a/Autopilot/AttitudeManager/SensorFusion.hpp
+++ b/Autopilot/AttitudeManager/SensorFusion.hpp
@@ -1,0 +1,33 @@
+/**
+ * Sensor Fusion - Converts raw sensor data into human readable formatted structs
+ * Author: Lucy Gong
+ */
+#ifndef SENSORFUSION_HPP
+#define SENSORFUSION_HPP
+
+//Include involved sensors
+#include "IMU.hpp"
+#include "airspeed.hpp"
+
+struct SFOutput_t{
+    float IMUroll, IMUpitch, IMUyaw; //in rad (for now)
+    float IMUrollrate, IMUpitchrate, IMUyawrate; //in rad/s (for now)
+
+    float Airspeed; //in m/s (for now)
+};
+
+struct SFError_t{
+    int errorCode;
+};
+
+/**
+ * Converts raw IMU output into roll, pitch and yaw angles
+ * 
+*/
+void IMUconv(IMUData_t *imu);
+
+
+SFError_t SF_GetResult(SFOutput_t *Output, SFError_t *Error);
+
+
+#endif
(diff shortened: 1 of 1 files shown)


avlepage (approved):

Autopilot/AttitudeManager/SensorFusion.hpp
```diff
@@ -0,0 +1,33 @@
+/**
+ * Sensor Fusion - Converts raw sensor data into human readable formatted structs
+ * Author: Lucy Gong
+ */
+#ifndef SENSORFUSION_HPP
+#define SENSORFUSION_HPP
+
+//Include involved sensors
+#include "IMU.hpp"
```
Fberbari: Why does the caller of your module need to know about these sensors ? Isn't your module responsible for internally talking to them ?

Autopilot/AttitudeManager/SensorFusion.hpp
```diff
@@ -0,0 +1,33 @@
...
+};
+
+/**
+ * Converts raw IMU output into roll, pitch and yaw angles
+ * 
+*/
+void IMUconv(IMUData_t *imu);
+
+
+SFError_t SF_GetResult(SFOutput_t *Output, SFError_t *Error);
```
Fberbari: I'm confused about the usage of your module. Do you expect the state machine to talk with the sensors, collect the IMU struct and hand it to your module for processing? If so, where does the state machine give your module the airspeed reading ?

That would be a fine implementation, but I was under the impression that we agreed your module would internally collect sensor info.

Fberbari: Also, why do we need 2 seperate functions to do this job ?

Fberbari: Also, your error struct is both passed in as an argument and also returned. I don't see why.

Fberbari (changes requested):

Autopilot/AttitudeManager/SensorFusion.hpp
heyitslu: you're right, the SensorFusion module should be talking to them internally by calling Sensor.GetResult() - will change the header file to reflect this

Autopilot/AttitudeManager/SensorFusion.hpp
heyitslu: The intention was to keep the IMU conversion separate so it would be easy to find visually in the file - SF_GetResult() calls IMUconv. Since I can achieve the same effect by both declaring and defining IMUconv in SensorFusion.c, I will move this fn there.

heyitslu: I...am not sure why I did that, will be removed

Fberbari (approved):
Yeah, this looks much better

PR #38, opened 2020-07-09
Battery status header & mock class

sahil-kale:


Autopilot/Inc/batterysensor.hpp
```diff
@@ -0,0 +1,25 @@
...
+ *  
+ **/
+
+#ifndef BATTERYSENSOR_HPP
+#define BATTERYSENSOR_HPP
+
+/**
+ * Starts the ADC for the battery sensor 
+ */
+virtual void initBatterySensor();
```
heyitslu: Wait why are these not in a BatterySensor class?

Autopilot/Inc/batterysensor.hpp
```diff
@@ -0,0 +1,25 @@
+/**
+ * Battery Sensor Header File
+ * 
+ * This will basically sense the voltage of the battery.
+ * 
+ * Author: Sahil Kale
+ *  
+ **/
```
sahil-kale, line 8: Test comment?

heyitslu (commented):
There's no battery sensor class defined

Autopilot/Inc/batterysensor.hpp
```diff
@@ -0,0 +1,25 @@
...
+virtual void initBatterySensor();
+
+/**
+ *  Returns the battery voltage as a float
+ * 
+ * @param battery The selected battery to return the voltage from
+ */
+virtual float getBatteryVoltage(int battery);
+
+#endif
```
Fberbari: make sure to always add 1 blank line at the end of your files. They're currently both missing one. Not a huge deal, but will produce warnings every once in a while. I recommend you just rig your text editor to automatically add a blank line on save.

Autopilot/Inc/batterysensor.hpp
```diff
@@ -0,0 +1,25 @@
...
+ * Starts the ADC for the battery sensor 
+ */
+virtual void initBatterySensor();
+
+/**
+ *  Returns the battery voltage as a float
+ * 
+ * @param battery The selected battery to return the voltage from
+ */
+virtual float getBatteryVoltage(int battery);
```
Fberbari: I could be wrong, but is voltage the only thing wer'e gathering ?

Autopilot/Test/Inc/batterysensor_mock.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef BATTERYSENSOR_MOCK_HPP
+#define BATTERYSENSOR_MOCK_HPP
+
+#include <gmock/gmock.h>
+
+#include "batterysensor.hpp"
+
+class MockBatterySensor : public batterySensor
```
Fberbari, line 8: This inherits from the class "batterySensor", which, to Lucy's point doesn't exist.

Fberbari (changes requested):

Autopilot/Inc/batterysensor.hpp
sahil-kale: Yup, although it's for 2 batteries apparently so I just specified as an int

Autopilot/Inc/batterysensor.hpp
sahil-kale: Sounds good!

Autopilot/Test/Inc/batterysensor_mock.hpp
sahil-kale: Sorry; got blocked by work - my understanding so far was that these were to be mock implementations and didn't require a class to be written, although that seems to be incorrect lol. Do you want me to write a batterySensor class then or just remove the inheritment?

Fberbari: I recommend googling some examples of google mock so you can understand what's going on.

But long story short, your interface should be an abstract class. During test time, the mock class derives from it, and during real operation, the actual driver class derives from it. Lucy has a full working example in her PR as well, take a look.

sahil-kale: Alright, I think i've done it right in the most recent commit

Autopilot/Src/batterySensor.c
```diff
@@ -0,0 +1,9 @@
+void initADC()
+{
+
+}
+
+float getBatteryVoltage()
```
Fberbari: This is not how a class gets implemented. https://stackoverflow.com/questions/10141079/how-do-i-split-classes-with-inheritance-up-into-different-h-cpp-files is the start of an explanation.

Also, since this is a c++ class, the file extension should be .cpp. This confused CMake in the past.


All this being said, I don't think you should bother with providing the blank implementation for the moment. Let the person actually implementing this module worry about that when the time comes.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "C_Cpp.default.configurationProvider": "vscode-wpilib"
+}
```
Fberbari: Did this file get in by mistake ?

Fberbari (changes requested):

.vscode/settings.json
sahil-kale: Yeah, I'll uncommit that lol

Autopilot/Src/batterySensor.c
sahil-kale: ok sounds good, I'll uncommit the file then

Fberbari (approved):
Looks fine to me!

heyitslu (approved):
Looks like you fixed the class issue - nice!

PR #39, opened 2020-07-18
Created SensorFusion unit tests

heyitslu:
Files changed:

**For testing purposes**
SensorFusion.hpp
SensorFusion.cpp
Test_SensorFusion.cpp

**Other**:
IMU_Mock.hpp
airspeed_Mock.hpp
IMU.hpp
airspeed.hpp
Basically changed Sensor_GetResult to pass by reference via & instead of pointer so SetArgReferee() could be used during units tests.


Autopilot/Test/Src/Test_SensorFusion.cpp
```diff
@@ -0,0 +1,138 @@
...
+	airspeedTestData.sensorStatus = -1;
+	IMUTestData.isDataNew = 1;
+	airspeedTestData.isDataNew = 1;
+
+	SFError_t error;
+	SFOutput_t output;
+
+	/********************DEPENDENCIES*******************/
+	/********************STEPTHROUGH********************/
+	EXPECT_CALL(imumock, GetResult(_))
```
Fberbari, line 84: I had made that dependency section for the EXPECT_CALL stuff. I think things are much clearer if they are separated from the step through. Call it anything you like though.

Fberbari (commented):
THIS IS AWESOME!!!! google mock has finally been tamed. Looks great to me. 1 super minor comment but that's it.

LOL this whole passing by reference in C++ stuff looks fucked to me. But yeah, it's def the right thing to do. Kudos on figuring it out. Starting to feel like an old man stuck in my C ways.

Fberbari (approved):

Autopilot/Inc/IMU.hpp
```diff
@@ -18,7 +18,7 @@ struct IMUData_t {
     float gyrx, gyry, gyrz; 
 
     bool isDataNew; 
-    int status; //TBD but probably 0 = SUCCESS, -1 = FAIL, 1 = BUSY 
+    int sensorStatus; //TBD but probably 0 = SUCCESS, -1 = FAIL, 1 = BUSY
```
sahil-kale, line 21: Curious, could this be an enum instead so that we can add in more states if required and keep it clean?

PR #42, opened 2020-08-14
Attitude manager State Machine

sahil-kale:
Attitude Manager State Machine


Autopilot/AttitudeManager/Attitude_States.cpp
```diff
@@ -0,0 +1,68 @@
...
+static attitudeManagerState failed(void);
+
+/***********************************************************************************************************************
+ * Code
+ **********************************************************************************************************************/
+void initAttitudeState()
+{
+    //Initialize sensor classes via smart pointer 
+    //Initialize PID class
+    PIDController PID;
```
Fberbari: Variables declared here will go out of scope as soon as initAttitudeState exits. You probably want to have them be file scope instead (or private if your'e going to use a c++ class model).

Autopilot/AttitudeManager/Attitude_States.cpp
```diff
@@ -0,0 +1,68 @@
...
+    //Initialize PID class
+    PIDController PID; 
+    //Initialize SendtoSafety Module
+    SendToSafety_Init();
+}
+
+static void pathManagerInput()
+{
+    PMCommands Commands;
+    int pmErrorState = PM_Commands(Commands); //dummy commands call for now?
```
Fberbari: not dummy calls. I want this module to be built with TDD.

Autopilot/AttitudeManager/Attitude_States.cpp
```diff
@@ -0,0 +1,68 @@
...
+    currentState = sendToSafetyMode;
+}
+
+static void sendToSafety()
+{
+    float channelOut = 0; //default value
+    int errorState = OutputMixing_Execute(); //requires arguements!!
+    SendToSafety_Execute(channelOut); //sends the instructions to safety. Dumb q: why is it not liking me passing in the float?
+    currentState = pathManagerInputMode;
+}
```
Fberbari: Mentioned this a few times before. You need to add a new line at the end of your files. Get your text editor to do it automatically. It seems trivial, but some compilers on some OS,s will complain non stop about it.

Autopilot/AttitudeManager/Attitude_States.hpp
```diff
@@ -0,0 +1,14 @@
+/**
+ *  Attitude State Machine Header
+ */
+#ifndef ATTITUDESTATES_HPP
+#define ATTITUDESTATES_HPP
+
+//Using static so we don't have to create an instance of Attitude States
```
Fberbari: not what static means. This would be true if we were dealing with a c++ class.
Having a stand alone function be static means its scope is only limited to its object file. No one outside your module would be able to call these functions; not what you want.

On a side note, using static is a great way to make a global variable file scope or to make helper functions that you have no interest in any other modules knowing about private to that module.

Autopilot/AttitudeManager/Attitude_States.hpp
```diff
@@ -0,0 +1,14 @@
...
+static void initAttitudeState();
+static void runAttitude();
+
+PIDController PID;
```
Fberbari: There is no need for this to be exposed in the header file.

Fberbari (commented):
It's a start, but there isn't much of substance here yet. Am hoping to see more progress soon. Most important is that there needs to be a unit test file that gets built up as this module gets built up.

PR #45, opened 2020-09-05
Attitude Manager State Transition States

sahil-kale:
State Transition successful.

Note that there is now a change to the architecture - The states are now called as follows:

- Fetch Instructions
- Sensor Fusion
- PID loop 'em
- Send to safety


Fberbari (approved):
Looks great

PR #47, opened 2020-09-20
Altimeter Driver

sahil-kale:
Converting to the MS5637 pressure sensor.


Autopilot/Inc/altimeter.h
```diff
@@ -7,11 +7,12 @@ extern "C" {
 
 #include "i2c.h"
 
-HAL_StatusTypeDef MPL3115A2_Init();
+HAL_StatusTypeDef M5637_Init();
```
heyitslu: * MS5637 oop lol

Autopilot/Src/altimeter.c
```diff
@@ -1,61 +1,48 @@
 #include "altimeter.h"
 
-#define MPL3115A2_ADDR 0x60 << 1
-#define MPL3115A2_I2C I2C2
+#define MS5637_ADDR 0x76 //Address for the MS5637. Needs to be shifted for write/read (ref:datasheet)
```
heyitslu: you can alternatively have MS5637_READ 0xED MS5637_WRITE 0xEC defined so you don't have to shift

Autopilot/Inc/altimeter.h
sahil-kale: :O

Autopilot/Src/altimeter.c
sahil-kale: Ya that would actually be better, ty

sahil-kale:
Would like to test before we merge

Autopilot/Inc/altimeter.h
```diff
@@ -7,12 +7,23 @@ extern "C" {
```
Fberbari: Why do we have a .h and a .hpp interface ?

Autopilot/Inc/altimeter.hpp
```diff
@@ -6,13 +6,16 @@
 #ifndef ALTIMETER_HPP
 #define ALTIMETER_HPP
 
+#include "i2c.h"
```
Fberbari: unless you need these includes for something in the header file, it's always better to include things just in the cpp file. That way we don't create a large dependency chain and we don't give the power to anyone calling your module to also call random i2c functions.

Autopilot/Inc/altimeter.hpp
```diff
@@ -35,4 +38,25 @@ class Altimeter{
         virtual void GetResult(AltimeterData_t *Data) = 0; //
 };
 
+class MS5637 : public Altimeter {
+    public:
+        void Init();
```
Fberbari, line 43: Let's just make this a constructor. Base Class should have an empty constructor I guess.

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,153 @@
...
+#define MS5637_PROM_C1 0xA2
+#define MS5637_PROM_C2 0xA4
+#define MS5637_PROM_C3 0xA6
+#define MS5637_PROM_C4 0xA8
+#define MS5637_PROM_C5 0xAA
+#define MS5637_PROM_C6 0xAC
+
+bool MS5637::isI2CBusDefined = false;
+
+static I2C_HandleTypeDef* hi2c;
```
Fberbari, line 31: any reason this isn't a private variable ?

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,153 @@
...
+   
+   int64_t P = ((D1*SENS/2097152-OFF)/32768); //Pressure, 1000 -> 120000
+
+   *rawPressure = P;
+   *rawTemperature = TEMP;
+   
+
+}
+
+ float MS5637::getPressure() {
```
Fberbari: ocd: open curly brace on the same line.

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,153 @@
...
+   std::time_t result = std::time(0);
+   return static_cast<int>(result);
+ }
+
+ void MS5637::Begin_Measuring()
+ {
+   altitudeMeasured = getAltitude();
+   pressureMeasured = getPressure();
+   temperatureMeasured = getTemperature();
+   dataIsNew = true;
```
Fberbari, line 123: remember that begin_measuring and get_result are called by 2 different threads. if you want dataIsNew, atltitudeMeasured... to be shared between them, they should be made static.

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,153 @@
...
+ {
+   std::time_t result = std::time(0);
+   return static_cast<int>(result);
+ }
+
+ void MS5637::Begin_Measuring()
+ {
+   altitudeMeasured = getAltitude();
+   pressureMeasured = getPressure();
+   temperatureMeasured = getTemperature();
```
Fberbari: Each one of these function calls eventually ends up setting up an I2C exchange to get the same information 3 times. I2C exchanges take a very long time compared to our processor speed, so it would be far less resource intensive to contact the chip just once and store the data needed for the math rather than get it 3 times from the sensor.

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,153 @@
...
+   dataIsNew = true;
+   timeOfResult = getCurrentTime();
+
+ }
+
+
+ void MS5637::GetResult(AltimeterData_t *Data)
```
Fberbari, line 129: The getResult function should not be contacting sensors. This function should return right away with data already fetched by Begin_measuring. If that data is still being fetched for some reason (which should never happen if we get our timing right), GetResult should just indicate that the data isnt ready and return.

The idea is that the getResult call is part of the state machine thread. We don't want to have to wait for sensors to respond one after the other in this thread. It would take a really long time and potentially mess up our timing.

Fberbari (changes requested):
Very solid. I2C and data processing stuff looks good so far as I can tell. The way getResult and BeginMeasuring work together needs some work though.

Autopilot/Src/altimeter.cpp
sahil-kale: OHHHH this is true

Autopilot/Src/altimeter.cpp
sahil-kale: Would it make sense to just make begin_measuring a static function then?

Autopilot/Src/altimeter.cpp
sahil-kale: Oh whoops, I forgot to update these to the new variables I added. Done :D

Autopilot/Inc/altimeter.h
sahil-kale: Oh, I forgot to ask. I suppose now that we're moving to .hpp, I'll get rid of this file lol

Autopilot/Src/altimeter.cpp
Fberbari: if it's static, then the constructor won't be called (so the I2C bus might not get setup).

Fberbari (commented):
CI build failed  as well

sahil-kale:
Strange... the testing framework seems to not line my code :/

sahil-kale:
Ok *should* work now

Autopilot/Inc/altimeter.hpp
```diff
@@ -35,4 +38,25 @@ class Altimeter{
...
+        void GetResult(AltimeterData_t *Data);
+    private:
+        uint32_t readFromMS5637(uint32_t commandToWrite);
+        void getRawPressureAndTemperature(int64_t *rawPressure, int64_t *rawTemperature);
+        uint32_t getCurrentTime();
+        float getTemperature();
+        float getPressure();
+        float getAltitude();
+        
+        static inline float altitudeMeasured, pressureMeasured, temperatureMeasured;
```
Fberbari: inline variables in a c++ 17 feature. At the moment, our build scripts only guarantee support of c++ 11.  This might be why your build is failing.

sahil-kale: D:

sahil-kale: @Fberbari Any clue on how to use a solution similar to inline? The only other way I could see us doing this is physically putting it at the top of the cpp file 😨

Fberbari: Yeah, that would be fine. Is there something wrong with putting them at the top of the file ? I thought it was pretty standard. For instance in attitudeStateClasses.cpp, in the state machine commit, this section appears at the top:

`/***********************************************************************************************************************
 * Definitions

float OutputMixingMode::_channelOut[4];
PMCommands fetchInstructionsMode::_PMInstructions;
SFOutput_t sensorFusionMode::_SFOutput;
PID_Output_t PIDloopMode::_PidOutput;`

Autopilot/Src/altimeter.cpp
sahil-kale: I'd have to include i2c.h in the altimeter.hpp file then for it to compile without errors.

Fberbari: t's fine to include things in the header so long as you have a reason to. This looks like a reason. We're just trying to prevent including files that serve no purpose.

But I don't really want to get hung up on that atm. I'll take a look at the rest. If things work, we can just move on.

PR #50, opened 2020-09-27
State machine clean

Fberbari:
With this PR, we've got a working state machine and its accompanied unit tests. Although some of the modules inside it (like PID) need some work, the state machine flow and data hand off functions correctly.

Changes that were made that don't seem obvious to the purpose of this commit are the following:

 Autopilot/CMakeLists.txt : Since so many new unit tests came in, I had to create multiple unit test executables (rather than 1 massive one). That way, different test suites could target exactly what they're interested in.

Autopilot/Tools/build.bash : Supports the above Cmake change.

Movement of the SensorFusion and OutputMixing unit test files: Better directory structure of the Test folderwas added to accommodate the above change


sahil-kale (approved):
Didn't see anything out of the ordinary :)

Autopilot/AttitudeManager/attitudeStateClasses.hpp
```diff
@@ -0,0 +1,127 @@
...
+        void exit(attitudeManager* attitudeMgr) {(void) attitudeMgr;}
+        static attitudeState& getInstance();
+    private:
+        FatalFailureMode() {}
+        FatalFailureMode(const FatalFailureMode& other);
+        FatalFailureMode& operator =(const FatalFailureMode& other);
+};
+
+
+//Reminder: add failed state mode?
```
heyitslu, line 127: Failed state mode was added already haha

heyitslu (approved):
Looks good to me!

Autopilot/AttitudeManager/attitudeStateClasses.hpp
Fberbari: good eye lmao, will fix next time this file is touched

PR #51, opened 2020-10-08
Fixed the PID module and added some unit tests to it

Fberbari:
The PID module now no longer relies on time. On top of this, it now allows the user the option to provide it with a measured derivative.

A couple of very simple unit tests were added for it and the attitude manager state machine was updated to use its new interface.


Autopilot/Src/PID.cpp
```diff
@@ -1,78 +1,72 @@
...
+ * Code
+ **********************************************************************************************************************/
 
-/* Generic PID functions. Can be used to PID other things (flaps, etc) */
+PIDController::PIDController(float _kp, float _ki, float _kd, float _i_max, float _min_output, float _max_output){
+	kp = _kp;
+	kd = _kd;
+	ki = _ki;
 
+	i_max = _i_max;
```
Pop0097, line 17: In PID.hpp i_max is declared with uint16_t, but the parameter "_i_max" is a float. This may result in truncation.

Autopilot/Inc/PID.hpp
```diff
@@ -0,0 +1,56 @@
...
+		* @param[in]	actualRate  The current measured derivative (This parameter is optional).
+		* @return					The result of the PID computation.
+		*/
+        float execute(float desired, float actual, float actualRate = std::nanf(""));
+
+    private:
+
+		float kp, kd, ki;
+		int16_t i_max;
+		float integral;
```
Pop0097: Not really an issue tbh, but the indentations could be aligned. Makes the code more readable.

Autopilot/Src/PID.cpp
Fberbari: good eye. Will do

Autopilot/Inc/PID.hpp
Fberbari: true. This has to do with the way GitHub displays tabs vs spaces. That file had a mix of both, but you're right. I'll just use one or the other so the indentations are consistent on any editor.

Pop0097: I didn't know GitHub did that. I'll be sure to keep an eye out for it too! :))

Autopilot/Src/PID.cpp
```diff
@@ -1,78 +1,71 @@
...
+	float derivative;
 
-  // check if we've gone too long without updating (keeps the I and D from
-  // freaking out)
-  if (delta_usec > PID_RESET_TIME || last_time == 0) {
-    delta_usec = 0;
-    integral = 0;
-    last_err = error;
-  }
+	integral +=  error;
```
sahil-kale, line 32: Would it make sense to add a flag to let whatever is calling the PID module to let it know that the system hasn't been updated for a while if this happens?

Autopilot/Inc/PID.hpp
```diff
@@ -0,0 +1,56 @@
+/**
+ * @file PID.hpp
+ * @author Ian Frosst
```
sahil-kale: Might wanna add your name too lol

Autopilot/Src/PID.cpp
Fberbari: I don't think so. This method only works properly if it's called at a constant interval (and the caller is warned about that in the .hpp file). If that doesn't happen, that's the caller's problem.

Autopilot/Inc/PID.hpp
Fberbari: fair enough :p

sahil-kale (approved):
Looks gucci

PR #52, opened 2020-10-15
IMU Driver

Pop0097:
IMU driver that collects data from a 3-dimensional accelerometer and gyroscope, and a temperature sensor. 

Uses SPI communication. 

Notes:
- Please make sure I implemented SPI properly (read, write, error catching, etc.)
- GPIOPin configurations. I don't know which settings we are using, so I put placeholders for now.

Hope you are having a nice reading week.


Autopilot/Inc/IMU.hpp
```diff
@@ -1,25 +1,75 @@
...
+//Sensor Sensitivities
+#define ACCEL_SENSITIVITY_4G 8192.0f
+#define GRYO_SENSITIVITY_500 65.5
+#define TEMPERATURE_SENSITIVITY 328.6
+
+/*** REGISTER DEFINITION BEGINS ***/
+
+//Configuration
+#define REG_CONFIG 0x1A
+#define REG_GYRO_CONFIG 0x1B
```
Fberbari: the header file is generic and could potentially belong to a bunch of different IMUs, anything specific to your IMU (like register names and particular sensitivities), belongs in the sourcefile.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
+/**
+ * IMU Sensor
+ * Author(s): Dhruv Rawat
+ */
+
+#include "IMU.hpp"
+#include "spi.h"
```
Fberbari: not needed.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include "SPI.hpp"
+#include "Status.hpp"
+#include "GPIO.hpp"
+ 
+bool ICM20602::isSPIBusDefined = false;
```
Fberbari: Standby about this. Sahil's working on a better method involving singletons that avoids all these sketchy global variables. Will let you know when he finishes it.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+         Data should be transitioned on the falling edge of SPC
+
+         From the above, we know that CPHA must be 0. 
+         Looking at the SPIPort::set_slave() method, CPOL is 1.
+
+         Thus, we should use SPI_MODE_2
+      */
+
+      hspi_1.mode = SPI_MODE_2;
+      hspi_1.master = false;
```
Fberbari: the microcontroller is in fact the master in this instance. THe IMU is the slave.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include "IMU.hpp"
+#include "spi.h"
+#include "SPI.hpp"
+#include "Status.hpp"
+#include "GPIO.hpp"
+ 
+bool ICM20602::isSPIBusDefined = false;
+bool ICM20602::dataIsNew = false;
+ 
+static SPI_HandleTypeDef* hspi1;
```
Fberbari: what's this ?

Also on line 22 you have a different variable with the same name.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+   }
+
+   accelConversionFactor = ACCEL_SENSITIVITY_4G;
+   gyroConversionFactor = GRYO_SENSITIVITY_500;
+   tempConversionFactor = TEMPERATURE_SENSITIVITY;
+}
+
+void ICM20602::write_data(uint8_t *writeData) { 
+   spi_port->set_slave(0); //Sets CS to HIGH
+   spi_port->exchange_data(writeData, 0x00, 1);
```
Fberbari: second argument is 0. But the prototype tells you it expects a pointer to a buffer. The exchange data method will try to fill up address 0x00 with data. If the compiler doesn't see this as an error, you'll be overwriting whatever stuff is in the lower addresses and cause horrible unexpected problems that will take weeks to figure out LOL. 

You'll still need to pass the address of an actual buffer of the correct size, even if you don't care about what it gets populated with.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+      write_data(writeData); 
+   }
+
+   accelConversionFactor = ACCEL_SENSITIVITY_4G;
+   gyroConversionFactor = GRYO_SENSITIVITY_500;
+   tempConversionFactor = TEMPERATURE_SENSITIVITY;
+}
+
+void ICM20602::write_data(uint8_t *writeData) { 
+   spi_port->set_slave(0); //Sets CS to HIGH
```
Fberbari: misunderstanding of the set_slave method. 
This method doesn't directly control the state of the CS pin, that's done internally by the exchange_data method. The point of this method is to tell the driver which slave you're trying to talk to, since it could be handling multiple other slaves on the same bus. So back when you called add_slave(), that method returns a number by which to identify the slave you just added, that's the number you should be passing into set_slave(), and you only need to do it once before each transaction.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+   //Store both High and Low Byte values
+   int16_t shiftedSensorAccX, shiftedSensorAccY, shiftedSensorAccZ, shiftedSensorTemp, shiftedSensorGyroX, shiftedSensorGyroY, shiftedSensorGyroZ; 
+
+   //Uses the sensor registers to get raw data for all sensors
+   *buffer = REG_ACCEL_XOUT_H | 0x80; 
+   
+   for(int i = 0; i < 14; i++) {
+      dataFromSensor[1] = 0x00; 
+      spi_port->set_slave(0); 
+      sensorSuccess = spi_port->exchange_data(buffer, dataFromSensor, 2);
```
Fberbari: buffer is of size 1. But you instruct exchange data to send/receive 2 bytes, on the first byte, it will use whatever's in the buffer, but on the second byte, it will overrun the buffer and start using data at some address that you don't own. This will sometimes cause problems and sometimes not, and be very hard to detect during debugging, so you should make sure that your data buffers are the correct size.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+   //Uses the sensor registers to get raw data for all sensors
+   *buffer = REG_ACCEL_XOUT_H | 0x80; 
+   
+   for(int i = 0; i < 14; i++) {
+      dataFromSensor[1] = 0x00; 
+      spi_port->set_slave(0); 
+      sensorSuccess = spi_port->exchange_data(buffer, dataFromSensor, 2); 
+      spi_port->set_slave(0); 
+      *buffer = REG_ACCEL_XOUT_H + 1; //Moves onto next register
```
Fberbari: I think you mean to increment the address by one every time ? The current way, the address will always be the same lool.

But yes, this would work, but take about twice as long as it'd need to. I've attached a diagram of what I mean.

image

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+   shiftedSensorAccY = (raw_data[2] << 8) + raw_data[3];
+   shiftedSensorAccZ = (raw_data[4] << 8) + raw_data[5];
+
+   shiftedSensorTemp = (raw_data[6] << 8) + raw_data[7];
+
+   shiftedSensorGyroX = (raw_data[8] << 8) + raw_data[9];
+   shiftedSensorGyroY = (raw_data[10] << 8) + raw_data[11];
+   shiftedSensorGyroZ = (raw_data[12] << 8) + raw_data[13];
+
+   //Converts 16-bit integers to a float. These are the actual measurements
```
Fberbari, line 196: All this arithmetic looks great, pretty cleanly done too.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+   *temp = ((float) shiftedSensorTemp)/tempConversionFactor;
+
+   *gyrx = ((float) shiftedSensorGyroX)/gyroConversionFactor;
+   *gyry = ((float) shiftedSensorGyroY)/gyroConversionFactor;
+   *gyrz = ((float) shiftedSensorGyroZ)/gyroConversionFactor;
+}
+
+uint32_t ICM20602::get_current_time() {
+   std::time_t currentTime = std::time(0);
```
Fberbari: I talked to Sahil about this, I think we're going to ditch this whole time thing. It's much harder than this to actually get time on a micro-controller and it's of very low value. dw about it right now though, We'll see what Sahil decides on first.

Fberbari (commented):
Good work so far

Autopilot/Src/IMU.cpp
Pop0097: Alright, I previously thought this was in the perspective of the sensor (asking if the sensor was the master). My bad. Will fix asap!

Autopilot/Src/IMU.cpp
Pop0097: Alright. I'll call spi_port->set_slave() before I call spi_port->exchange_data().

Autopilot/Src/IMU.cpp
Pop0097: Clarification:

If I pass an array with 15 elements into the exchange_data() method for the rx_buffer parameter (the first element will be the dummy byte), will I get the sensor data from the next 14 registers in the following 14 elements of the array? 

Also the value I pass into the tx_buffer parameter only needs to be a uint8_t value that is the register I want to start measuring from (in this case, X_OUT_H), right? I don't need to pass in a 15 element array that includes the dummy bytes in MISO? 

In my most recent commit, I structured the code using what I described above. (in case I explained it poorly)

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+   accelConversionFactor = ACCEL_SENSITIVITY_4G;
+   gyroConversionFactor = GRYO_SENSITIVITY_500;
+   tempConversionFactor = TEMPERATURE_SENSITIVITY;
+}
+
+void ICM20602::write_data(uint8_t *writeData) { 
+   uint8_t *imuPlaceholderByte = 0x00;
+   spi_port->set_slave(imuSlaveIdentifier);
+   spi_port->exchange_data(writeData, imuPlaceholderByte, 2);                                                                    //WRITEDATA ARRAY HAS TWO ELEMENTS, THUS THE imuBuffer SIZE IS TWO HERE!!!!
```
Pop0097: Is the buffer size (the integer 2) the correct value? I am passing in writeData, which has **two** uint8_t elements, as the tx_data parameter.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+      //Verify chip is working properly
+      *imuBuffer = REG_WHO_AM_I | 0x80;
+      uint8_t dataFromIMUSensor[2] = {0x00, 0x00}; // dataFromSensor[0] -> dummy byte (contains SPI address); dataFromSensor[1] -> data byte
+      spi_port->set_slave(imuSlaveIdentifier);
+      sensorSuccess = spi_port->exchange_data(imuBuffer, dataFromIMUSensor, 1);
+
+      if(sensorSuccess != STATUS_CODE_OK) {
+         //Debug code here
+      }
```
Pop0097: Is there a specific way we would like error handling to be done?

Fberbari: getResult should return an error code if something goes wrong. dw about adding debug code in the constructor for the moment.

Autopilot/Src/IMU.cpp
Pop0097: THING ABOUT RX IS CORRECT

THING ABOUT TX IS INCORRECT, SEND IN 15 ELEMENT ARRAY!! FIRST ELEMENT IS THE X_OUT_H AND THE REST ARE DUMMY BYTES

Pop0097: Also the size_t parameter should be 15

Autopilot/Src/IMU.cpp
Pop0097: THE imuPlaceholderByte should be a 2 element uint8_t array. 

The length parameter is correct

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+ 
+void ICM20602::Begin_Measuring() {
+   get_accel_temp_gyro_reading(&measuredAccX, &measuredAccY, &measuredAccZ, &measuredGyroX, &measuredGyroY, &measuredGyroZ, &measuredTemp);
+   dataIsNew = true;
+}
+ 
+void ICM20602::GetResult(IMUData_t &Data) {
+   dataIsNew = false;
+   if (sensorSuccess == STATUS_CODE_OK) {
```
Fberbari, line 214: there's a bug here. if I call GetResult twice in a row, it will tell me data is new both times rather than telling me it's old the second time.

Autopilot/Src/IMU.cpp
sahil-kale: Singleton PR is up, after it's approved this should be ported into that kind of driver structure

Autopilot/Inc/IMU.hpp
```diff
@@ -1,25 +1,31 @@
...
 struct IMUData_t {
 
-    float magx, magy, magz;
+    float magx, magy, magz; //ICM20602 does not have a magnetometer. Thus, these readings will be left as NAN
     float accx, accy, accz;
     float gyrx, gyry, gyrz; 
+    float temp; 
 
     bool isDataNew; 
     int sensorStatus; //TBD but probably 0 = SUCCESS, -1 = FAIL, 1 = BUSY
```
sahil-kale, line 28: More of a quirk, but I think an enum would be better suited for states than integers defined in comments imo

sahil-kale (changes requested):
Looks good. I think all that's left is to make it return a singleton object and that should be it

Autopilot/Inc/IMU.hpp
Pop0097: We'll have to decide on a consistent structure for this before we change this so all drivers return similarly structured data.

Pop0097:
The tests are failing because of the following error: 

Screen Shot 2020-10-21 at 8 56 46 PM

Autopilot/Inc/IMU.hpp
```diff
@@ -45,23 +47,42 @@ class IMU{
...
+    private:
+        //Constructor 
+        ICM20602();
+
+        //Object Instance
+        static ICM20602* imu_Instance;
+
+        //Variables
+        uint32_t timeOfResult;
+        static bool isSPIBusDefined;
```
sahil-kale: no need for this anymore :D

sahil-kale (changes requested):
Looks good! Just need to take out the static bool thingy since there'll only ever be one object lol

Autopilot/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ if (${KIND_OF_BUILD} STREQUAL "FOR_TARGET")
           -iex "disconnect" -iex "quit")
 
 elseif(${KIND_OF_BUILD} STREQUAL "UNIT_TESTS")
```
Fberbari, line 89: In the future,  it's much better to merge the PR stuff into your branch (or rebase your branch onto the new PR). Copying changes manually like this might cause merge conflicts when the time comes.

Autopilot/Src/IMU.cpp
```diff
@@ -0,0 +1,234 @@
...
+      Data.gyrz = measuredGyroZ;
+      Data.magx = NAN;
+      Data.magy = NAN;
+      Data.magz = NAN;
+      Data.temp = measuredTemp;
+      Data.sensorStatus = 0;
+   } else { //In case data was not received. Stick with old data
+      Data.isDataNew = false;
+      Data.sensorStatus = -1; 
+   }
```
Pop0097, line 231: Note to future Dhruv when debugging the driver:

See if making sensorStatus -1 is needed. If the status is not 0, AttitudeManager will go into "fatal failure mode" which is NOT GOOD.

PR #53, opened 2020-10-18
Fixed "undefined vtable" bug

Fberbari:
Because the implementations of certain abstract classes where not
present during unit testing (hardware drivers), certain linkers on
certain platforms threw the undefined vtable error, even though the
methods were never called. The solution to this involved making sure the
corresponding mock classes were always around. This way, there is always
an implementation to be found. In practice, this involved adding a
"UNIT_TESTING" flag to the build and using preprocessor macros to
include the mock classes into the correct header files when needed.


Pop0097 (approved):
Looks good.

PR #54, opened 2020-10-21
Altimeter driver: Singleton Style

sahil-kale:
Okay so with my rather crude knowledge of singletons, I've changed the altimeter driver.

If I'm not mistaken, the altimeter should now be called like so:
`MS5637::GetInstance()->Begin_Measuring()`

Here are the tutorials I used:
http://www.yolinux.com/TUTORIALS/C++Singleton.html
https://www.youtube.com/watch?v=PPup1yeU45I


.gitignore
```diff
@@ -55,6 +55,9 @@ mx.scratch
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# VSCode
+.vscode
```
Pop0097, line 59: I think you need to remove the viscose files and all. Don't want them getting merged with the rest of the repo.

Autopilot/Inc/altimeter.hpp
```diff
@@ -35,4 +38,24 @@ class Altimeter{
...
 
+class MS5637 : public Altimeter {
+    public:
+        MS5637(const MS5637*) = delete; //Apparently if you try to copy a singleton this will give you errors?
+        static MS5637* GetInstance();
+        void Init();
+        void Begin_Measuring();
+        void GetResult(AltimeterData_t *Data);
+    private:
+        MS5637() {} //Constructor can never be called muwhahaha
```
Pop0097: Do we also want to make a private assignment operator? (Just to prevent mistakes in other parts of the code that depend on the drivers).

Example cited in the YoLinux.com tutorial:

> private:
> Logger& operator=(Logger const&){};  // assignment operator is private

Autopilot/Test/Src/Test_Altimeter_Skeleton.cpp
```diff
@@ -1,7 +1,7 @@
 #include <gtest/gtest.h>
 #include "fff.h"
 #include "i2c.h"
-#include "altimeter.h"
+//#include "altimeter.h"
```
Pop0097: You can probs just delete this since you deleted the altimeter.h file.

Pop0097 (commented):
Singleton implementation looks good.

Autopilot/Src/MS5637.cpp
```diff
@@ -0,0 +1,143 @@
...
+  if(!s_Instance) //Generates new instance if one doesn't exist :O
+  {
+    s_Instance = new MS5637;
+  }
+
+  return s_Instance;
+}
+
+
+ void MS5637::Init()
```
Fberbari: Let's ditch the init function and put this in the constructor instead.

Autopilot/Inc/altimeter.hpp
```diff
@@ -35,4 +38,24 @@ class Altimeter{
...
+        static MS5637* s_Instance;
+        uint32_t readFromMS5637(uint32_t commandToWrite);
```
Fberbari, line 45: A function called "read..." takes as a parameter a variable name "CommandToWrite" is a bit confusing

.gitignore
sahil-kale: The .gitignore means that it will ignore any files inside the .vscode directory and you (or anyone else actually) don't have to worry about accidentally committing vscode files lol
- Note: this only works if the file isn't cached

Autopilot/Inc/altimeter.hpp
sahil-kale: hmmm ya probs would be good

sahil-kale: Actually I think a reference to the instance should be okay, as long as you don't do something like
`MS5637 reference = MS5637::GetInstance->function()`
instead of
`MS5637* reference`

should be okay since line 43 takes care of that edge case and will not compile lol

Fberbari (commented):
I'm really happy with this! We're not gonna merge it until it's tested on the bench, but I'd say it's ready for that at this point.

Autopilot/Src/MS5637.cpp
sahil-kale: Okay, should we keep init() in the base class as well for now or no?

Fberbari: na get rid of it as well. Having it there is the most OOP noob thing I've ever done and I don't want to look at it anymore lmaooo.

sahil-kale: LOL noted

Autopilot/Inc/altimeter.hpp
sahil-kale: changed to commandToSend

Pop0097 (commented):
Looks good to me.

PR #56, opened 2020-10-31
GPS driver

Pop0097:
The GPS driver for the NEOM8 GPS Module.

### Some notes
- **This PR also contains older code from my IMU Driver PR**. Fixing it is being a pain, so I just deleted the files from the branch. 

- A lot of the code, parsers, etc. were translated from [PicPilot](https://github.com/UWARG/PICpilot/tree/staging/Autopilot/GPS). Files that came in handy include utilities.c and MTK33x9.c.

- I added some changes to the UART wrapper. My first change enabled me to initialize the function pointer member in the DMAConfig structure. The goal is that after the DMA interrupt is triggered, the desired function (handle_DMA_input() in gps.cpp) is called. The second change I did was create a getter function in UARTPort so I can access the double edged queue, rx_queue, parameter. I needed access to it so I could transfer the values in memory to the gps buffers in handle_DMA_input(). 

- There are some huge assumptions regarding UART. I have commented them in the handle_DMA_input() method. 

- I tested the parsers to ensure they worked. I will do more testing to ensure they are 100% fine!

- I switch from using bit specific integer declaration (ex. using uint8_t and uint16_t) to using "int" in the code. I did this because the parsers were not working in CLion when I used the bit-specific integer datatypes. If need be, I may convert all integer declarations to "int". Please advise on this.

Stay safe,

Dhruv R


Pop0097:
IMU files are included for some reason. Will need to remove them.

Autopilot/Inc/IMU.hpp
```diff
@@ -1,71 +0,0 @@
-/**
```
Fberbari, line 1: we delete this file ? Is it affecting gps somehow ?

Autopilot/Inc/gps.hpp
```diff
@@ -1,22 +1,15 @@
...
  * Gps sensor interface and part number selector.
- * Author: Anthony Berbari
+ * Author(s): Anthony Berbari, Dhruv Rawat
  */
 
 #ifndef GPS_HPP
 #define	GPS_HPP
 
 #include <stdint.h>
+#include "stm32f7xx_hal.h"
```
Fberbari: headers should never include hal stuff. If they do, anything that uses them becomes impossible to unit test. You should include this in your cpp file instead.

Autopilot/Inc/gps.hpp
```diff
@@ -55,7 +42,56 @@ class Gps
...
+		//Variables
+		bool isDataNew = false;
+		long double measuredLatitude, measuredLongitude;  // 8 Bytes
+		float measuredUtcTime;     // 4 Bytes. Time in seconds since 00:00 (midnight)
+		float measuredGroundSpeed; // in m/s
+		int measuredAltitude; // in m
+		uint16_t measuredHeading; // in degrees. Should be between 0-360 at all times, but using integer just in case
+		uint8_t measuredNumSatellites;  // 1 Byte
+
+		//Methods (Some code was transferred from PicPilor)
```
Fberbari: typo, picpilot***

Autopilot/Libraries/Drivers/Src/stm32f7/UART.cpp
```diff
@@ -406,6 +406,25 @@ StatusCode UARTPort::resetDMA() {
 	return STATUS_CODE_OK;
 }
 
+StatusCode UARTPort::registerDMAReceiveCallback(void (*f)()) {
```
Fberbari, line 409: good work on understanding wtf is going on and being able to fix it. I don't really know what's going on, but that's ok. Will revisit once we can test this driver on the hardware.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
+/**
+ * Gps sensor
+ * Author(s): Dhruv Rawat
+ */
+
+#include "gps.hpp"
+#include "UART.hpp"
+#include "../Libraries/STM32F7xx_HAL_Driver/Src/stm32f7xx_hal_uart.c"
```
Fberbari: Why are we including .c files ?? That doesn't seem right. Also, this file is very deep inside the hal layer, I'd imagine you'd never have to touch it. Should uart.hpp not be enough ?

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+#define GGA_MESSAGE "GPGGA"
+#define VTG_MESSAGE "GPVTG"
+
+NEOM8* NEOM8::gps_Instance = NULL;
+UARTPort *gpsUART;
+
+char vtgValues[GPS_UART_BUFFER_SIZE];
+char ggaValues[GPS_UART_BUFFER_SIZE];
+char generalBuffer[GPS_UART_BUFFER_SIZE];
```
Fberbari: I'd be much better to have these variables be private member variables than be global like this.

If you really want them out here though, make sure they're declared static. That way, they're file scope instead of global.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+uint8_t tx_buffer[GPS_UART_BUFFER_SIZE];
+uint8_t rx_buffer[GPS_UART_BUFFER_SIZE];
+
+bool isGPSConfigured = false;
+bool dataAvailable = false;
+bool newGGAData = false;
+bool newVTGData = false;
+
+void handle_DMA_input();
```
Fberbari: functions declared here should be static so that they are file scope

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+    //GPS Configuration
+    gpsUART->transmit((uint8_t*) PUBX_CONFIG_NMEA, sizeof(PUBX_CONFIG_NMEA));
+    //Turns off receivers (resets them)
+    gpsUART->transmit((uint8_t*) PUBX_SET_GGA_OFF, sizeof(PUBX_SET_GGA_OFF));
+    gpsUART->transmit((uint8_t*) PUBX_SET_VTG_OFF, sizeof(PUBX_SET_VTG_OFF));
+    //Turns on receivers
+    gpsUART->transmit((uint8_t*) PUBX_SET_GGA, sizeof(PUBX_SET_GGA));
+    gpsUART->transmit((uint8_t*) PUBX_SET_VTG, sizeof(PUBX_SET_VTG));
+    
+    isGPSConfigured = true;
```
Fberbari: I don't see the point of this variable.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+    gpsUART->registerDMAReceiveCallback(handle_DMA_input);
+    //GPS Configuration
+    gpsUART->transmit((uint8_t*) PUBX_CONFIG_NMEA, sizeof(PUBX_CONFIG_NMEA));
+    //Turns off receivers (resets them)
+    gpsUART->transmit((uint8_t*) PUBX_SET_GGA_OFF, sizeof(PUBX_SET_GGA_OFF));
+    gpsUART->transmit((uint8_t*) PUBX_SET_VTG_OFF, sizeof(PUBX_SET_VTG_OFF));
+    //Turns on receivers
+    gpsUART->transmit((uint8_t*) PUBX_SET_GGA, sizeof(PUBX_SET_GGA));
+    gpsUART->transmit((uint8_t*) PUBX_SET_VTG, sizeof(PUBX_SET_VTG));
```
Fberbari: Config seems correct. Nicely done.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+    
+    isGPSConfigured = true;
+}
+
+void handle_DMA_input() {
+    if (!isGPSConfigured) {
```
Fberbari: no point in this, it can only be called after the constructor is called anyway, so the gps is guaranteed configured, right ?

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+        //Assumed the wrapper will handle clearing the registers and all.
+        return;
+    }
+
+    bool currentlyParsing = false;
+    uint16_t bufferIndex = 0;
+    uint8_t data = 0;
+
+    std::deque<uint8_t> dma_data = gpsUART->get_rx_queue(); //Copys double eneded queue in UARTPort class that contains data in the DMA registers
```
Fberbari, line 56: doesn't seem right. The dma should be moving the data for you. You should never have to interact with the uart data registers.

Autopilot/Inc/gps.hpp
```diff
@@ -55,7 +42,56 @@ class Gps
...
+class NEOM8 : public Gps 
+{
+	public:
+		NEOM8(const NEOM8*) = delete;
+		NEOM8* GetInstance();
+
+		/**
+         * Triggers interrupt for new GPS measurement - stores raw data in variables and returns right away
+         * */
+		void BeginMeasuring();
```
Fberbari: looks like this sensor sends data over by itself, so we never actually need to call this. Cool. You can write it as void BeginMeasuring(){}; and tell the user to not bother calling it.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,386 @@
...
+        intConverted = toConvert - 48;
+    } else if (toConvert >= 65 && toConvert <= 70) {
+        intConverted = toConvert - 0x37;
+    }
+
+    return intConverted;
+}
+
+//Checks if there is new data, validates it, and then converts it to a form that the autopilot will understand
+void NEOM8::get_gps_data() {
```
Fberbari: I think you forgot to actually return the data lol.

Autopilot/Inc/gps.hpp
```diff
@@ -55,7 +42,56 @@ class Gps
...
+
+		 /**GetResult should:
+         * 1. Reset dataIsNew flag
+         * 2. Transfers raw data from variables to struct
+         * 3. Updates utcTime and status values in struct as well
+         * */
+		void GetResult(GpsData_t &Data);
```
Fberbari: this function seems to not be implemented, there is a function called get_gps_data, but it doesent give anything back to the caler. Not sure what's going on.

Fberbari (commented):
Clearly, a lot of work went into it, well done. A lot of it is over my head for the moment. But it *generally* looks very good. What we'll do is make sure everything compiles and debug it on the hardware until it works. Then we can merge. Until then, you're free to put this aside and work on other stuff

Autopilot/Inc/IMU.hpp
Pop0097: Oh, when I made the gps branch, it included the past changes I made to the IMU files. So basically It just had outdated changes in the IMU PR. 

I deleted the files so we don't merge them by accident.

Autopilot/Src/gps.cpp
Pop0097: True, will remove.

Autopilot/Inc/gps.hpp
Pop0097: HUH!? I swear I wrote it. Must've accidentally deleted it.

Autopilot/Src/gps.cpp
Pop0097: I don't need to return the data from this function because in the parse_register() methods it stores the values in private variables of the NEOM8 instance. 

I must've deleted the GetResult() method by accident. Dk when it happened but it did :((

Autopilot/Inc/gps.hpp
Pop0097: Something needs to call get_gps_data() in order for the NMEA strings to be parsed. We may need this method to start parsing the data in the buffers... 

Unless I can call get_gps_data() from handle_DMA_input(). Tricky though since get_gps_data() is a member of the NEOM8 class while handle_DMA_input() is not...

Autopilot/Src/gps.cpp
Pop0097: Will need to explore more. I just called this so I could actually interact with the data in the loop. 

Without this, I didn't have anything concrete to define the size of the loop, etc.

Autopilot/Inc/gps.hpp
Fberbari: How about you store the raw data, then parse it at the time GetResult is called.

Pop0097: That's a great idea! Will do that!

Autopilot/Src/gps.cpp
Pop0097: Will keep them as static so I can access them in the handle_DMA_input() function.

Pop0097:
It turns out I needed to include the IMU.hpp file for the program to compile. We will need to remove it before merging this PR as the up-to-date IMU.hpp file is in PR 52.

Pop0097:
Note to self: Add GPS fix to the GPSData_t structure 

Also it is the first Linux test that is failing again :cries:. I will have to copy and paste the fix over from the IMU PR.

Fberbari (commented):
good stuff, I don't have any more comments. I'll say this is ready for bench testing.

As far as your workflow goes, try this next time:

Make a branch called IMU off of devel, do your work there, then make another branch called gps off of devel and do your work there. 

I think here you've made your gps branch off of IMU (right ?) you can make things easier for yourself by debasing the gps branch to the devel branch. But anyway, don't waste your time on that now.

PR #57, opened 2020-11-01
Stopped including Hal header files in our unit test build.

Fberbari:
This solves all kinds of sketchy problems that arose from including files that were specificaly meant for another hardware architecture. Previous to this, the fix was to compile with -fpermissive and hope for the best, though things started breaking in development (an issue found by Dhuruv while developing the gps driver). With this change, the test build will be far more robust. The trade off is that we won't be able to unit test any modules that call hal layer functions (like divers).


Pop0097 (approved):
The changes look good to me!

PR #58, opened 2020-11-04
Safety changes

sahil-kale:
Changed the way PWM is given to safety.

- Now, anything can call send to safety (i.e. camera gimbal, output mixing, etc) as much as it needs without having to account for anything else

### DIFF
--- a/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
+++ b/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
@@ -18,6 +18,8 @@ SendToSafety_error_t SendToSafety_Execute(int channel, int percent)
 {
     SendToSafety_error_t error;
     error.errorCode = 0;
+    //TODO: Add in something that can update the error code if required
+    //THIS IS TOP LEVEL PRIORITY ONCE INTERCHIP IS WRITTEN!!
     pwmPercentages[channel] = percent;
 
     Interchip_SetPWM(pwmPercentages);
--- a/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
+++ b/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
@@ -2,8 +2,8 @@
 #include "Interchip_A.h"
 #define PWM_CHANNELS 12
 
-int16_t pwmPercentages[PWM_CHANNELS] = {0};
-int16_t initialPWMPercentages[PWM_CHANNELS] = {0}; //TODO: put in initial PWM states in here. 
+static int16_t pwmPercentages[PWM_CHANNELS] = {0};
+static int16_t initialPWMPercentages[PWM_CHANNELS] = {0}; //TODO: put in initial PWM states in here. 
 
 
 void SendToSafety_Init(void)
--- a/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
+++ b/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
@@ -1,5 +1,5 @@
 #include "SendInstructionsToSafety.hpp"
-#include "interchip_A.h"
+#include "Interchip_A.h"
 #define PWM_CHANNELS 12
 
 int16_t pwmPercentages[PWM_CHANNELS] = {0};
--- a/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
+++ b/Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
@@ -1,0 +1,24 @@
+#include "SendInstructionsToSafety.hpp"
+#include "interchip_A.h"
+#define PWM_CHANNELS 12
+
+int16_t pwmPercentages[PWM_CHANNELS] = {0};
+int16_t initialPWMPercentages[PWM_CHANNELS] = {0}; //TODO: put in initial PWM states in here. 
+
+
+static void SendToSafety_Init(void)
+{
+    for(int i = 0; i < PWM_CHANNELS; i++)
+    {
+        pwmPercentages[i] = initialPWMPercentages[i];
+    }
+}
+
+static SendToSafety_error_t SendToSafety_Execute(int channel, int percent)
+{
+    SendToSafety_error_t error;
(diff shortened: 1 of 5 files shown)


Autopilot/AttitudeManager/SendInstructionsToSafety.hpp
```diff
@@ -31,9 +32,10 @@ void SendToSafety_Init(void);
 * Communicates the desired actuator commands encoded in the channelOut array over I2C to
 * the safety chip. Internally, this function performs a conversion from percentages to pwm values.
 * This function is non blocking and returns right away.
-* @param[in]		channelOut 		the array (size of 4 floats) of percentages each channel should be set to.
+* @param[in]		channelOut 		the output channel for PWM
```
Pop0097: the variable identifier is channel

Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
```diff
@@ -0,0 +1,25 @@
...
+    }
+}
+
+SendToSafety_error_t SendToSafety_Execute(int channel, int percent)
+{
+    SendToSafety_error_t error;
+    error.errorCode = 0;
+    pwmPercentages[channel] = percent;
+
+    Interchip_SetPWM(pwmPercentages);
```
Pop0097, line 25: I don't see a place where error.errorCode is changed if an error occurs.  Am I missing something?

Pop0097 (commented):
Code seems alright to me. Wait for Anthony to take a look tho because he'll know more about how this PR will integrate with the existing code.

Autopilot/AttitudeManager/SendInstructionsToSafety.hpp
sahil-kale: whoops you right lol

Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
sahil-kale: Yeah, this is a todo since interchip hasn't been written yet. I'll add it in a comment for now I guess. If interchip fails, it should return a status code, and that will feed into the error struct is the current plan iirc

Autopilot/AttitudeManager/SendInstructionsToSafety.cpp
```diff
@@ -0,0 +1,27 @@
+#include "SendInstructionsToSafety.hpp"
+#include "Interchip_A.h"
+#define PWM_CHANNELS 12
+
+int16_t pwmPercentages[PWM_CHANNELS] = {0};
+int16_t initialPWMPercentages[PWM_CHANNELS] = {0}; //TODO: put in initial PWM states in here.
```
Fberbari: THese variables should be declared static, so as to avoid them being global to all files.

Autopilot/AttitudeManager/SendInstructionsToSafety.hpp
```diff
@@ -31,9 +32,10 @@ void SendToSafety_Init(void);
 * Communicates the desired actuator commands encoded in the channelOut array over I2C to
 * the safety chip. Internally, this function performs a conversion from percentages to pwm values.
 * This function is non blocking and returns right away.
-* @param[in]		channelOut 		the array (size of 4 floats) of percentages each channel should be set to.
+* @param[in]		channel 		the output channel for PWM
+* @param[in]		percent 		the percentage between 0-100 that the PWM channel should be set to
 * @return							the error struct, containing all info about any errors that may have occured.
 */
-SendToSafety_error_t SendToSafety_Execute(float *channelOut);
+SendToSafety_error_t SendToSafety_Execute(int channel, int percent);
```
Fberbari, line 39: Is the idea to initiate an I2C exchange every time any channel is updated ? Or are you thinking about storing the variables inside the module until all channels are ready then shipping them off all at once.

Autopilot/Src/Interchip_A.c
```diff
@@ -31,8 +31,13 @@ void Interchip_Run(void const *argument) {
 
 // Public Functions to get and set data
 
-int16_t *Interchip_GetPWM(void) { return dataRX->PWM; }
-void Interchip_SetPWM(int16_t *data) {
+int16_t *Interchip_GetPWM(void)
```
Fberbari, line 34: Are these just cosmetic changes ?

Autopilot/AttitudeManager/SendInstructionsToSafety.hpp
Fberbari, line 39: why make percent an integer ? It's not a big deal, but to me it looks like we're sacrificing some resolution for no upside.

sahil-kale: Interchip currently uses int's instead of floats since PWM manager uses int's. I could probably just make it take in a float and then round it if you'd like?

Autopilot/Src/Interchip_A.c
sahil-kale: Yup

Autopilot/AttitudeManager/SendInstructionsToSafety.hpp
sahil-kale: So the way I understand interchip is that the I2C exchange is 'threaded', and the setPWM method just updates an array that stores the channel percentages. So safety and autopilot do their exchanges completely irrespective to when setPWM is called, which is smart cause then you don't have to wait for say grabber output when the PID output mode sends commands to the ailerons or something like that

sahil-kale: *Oh, the I2C exchange basically just sends this array full of channel percentages inside a struct lol

Fberbari: Oh that makes a ton of sense and is super clean. Sweet!

Fberbari: understood, thanks. Keep it the way it is then!

Fberbari (approved):

PR #59, opened 2020-11-09
Airspeed PR

sahil-kale:


SnackkOverflowError:
So, I think everything is good, except that I couldn't find an ADC specifically for airspeed, so I just used hadc2 instead.

Also, Im not sure if init() is actually needed, because the altimeter driver didnt use init afaik, but airspeed did.

Autopilot/Inc/airspeed.hpp
```diff
@@ -84,6 +86,26 @@ class dummyairspeed: public airspeed{
 
 };
 
+class MPXV7002DP : public airspeed {
+    public:
+        MPXV7002DP(const MPXV7002DP*) = delete;
+        static MPXV7002DP* GetInstance();
+        void GetResult(airspeedData_t &Data);
```
sahil-kale: This should be a pointer in case the struct is null lol

Autopilot/Inc/airspeed.hpp
```diff
@@ -84,6 +86,26 @@ class dummyairspeed: public airspeed{
...
+        void Begin_Measuring();
+        void Init();
```
sahil-kale: Everything in init can be put in the constructor

Autopilot/Inc/airspeed.hpp
```diff
@@ -84,6 +86,26 @@ class dummyairspeed: public airspeed{
...
+    private:
+        MPXV7002DP();
+        static MPXV7002DP* s_Instance; //single instance of airspeed
+        //command to interface with MPXV7002
+        double readFromMPXV7002DP();
+        int offset = 0;
```
sahil-kale: If I understand the code correctly could this be renamed to something like "zeroVelocityOffset" or something?

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,65 @@
...
+        sensorStatus = 0;
+}
+
+void MPXV7002DP::Init() {
+
+    //do I need to init the adc?
+
+    uint16_t adcVal = HAL_ADC_GetValue(&hadc2);
+    offset = adcVal - (ADC_RESOLUTION-1)/2;
+}
```
sahil-kale: See above comment about the init function

sahil-kale (commented):
So I can't request changes on this since I'm technically the PR author LOL

Autopilot/Inc/airspeed.hpp
Fberbari: yep. Please remove the init function both from the base and derived classes.

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <cmath>
+
+#define VS 5
+#define DENSITY_AIR 1.204
+#define ADC_RESOLUTION 4096
+
+
+MPXV7002DP::MPXV7002DP() {
+    //constructor
+    MX_ADC2_Init();
```
Fberbari: A quick look through adc.c shows me adc3 is used for airspeed. adc 2 is actually used for battery stuff. So you'll have to use adc3 in your driver.

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,65 @@
...
+void MPXV7002DP::Begin_Measuring() {
+        airspeed = readFromMPXV7002DP();
+        timeOfLastResult = getCurrentTime();
+        dataNew = true;
+        sensorStatus = 0;
+}
+
+void MPXV7002DP::Init() {
+
+    //do I need to init the adc?
```
Fberbari: all initialization is done by the MX_adc_init function, no need to do anything else

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    uint16_t adcVal = HAL_ADC_GetValue(&hadc2);
```
Fberbari: I think you're trying to get a zero speed reading right ?

If that's the case, I'd recommend taking multiple readings (maybe 100 ? ) and using a median or average or something.

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+        return 0;
+    }
+
+}
+
+uint32_t MPXV7002DP::getCurrentTime() {
+    std::time_t result = std::time(0);
+    return static_cast<int>(result);
+}
```
Fberbari: leave a blank line at the end of all code files. This complies with some unix rules. Is a bit nitpicky, but it might supress a warning some day, so may aswell get in the habit of doing it

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,65 @@
...
+    uint16_t adcVal = HAL_ADC_GetValue(&hadc2);
+    offset = adcVal - (ADC_RESOLUTION-1)/2;
+}
+
+
+double MPXV7002DP::readFromMPXV7002DP() {
+    uint16_t adcVal = HAL_ADC_GetValue(&hadc2) - offset;
+
+    if(adcVal < ADC_RESOLUTION/2) {
+        return -std::sqrt( (-10000*(adcVal/(ADC_RESOLUTION-1) - 0.5))/DENSITY_AIR );
```
Fberbari: I'm confused/there are some problems with the math here:
- What does 10000 mean ? I'd recommend using a define so people can understand what your'e doing.
- adcVal and ADC_RESOLUTION are both integers, so the compiler will do integer division and that division will be 0 every time. You can test this on some online cpp compiler. You'l have to cast one of the values to float or double to get the result you want.
- adcVal and ADC_resolution are always positive, right ? So won't the result of this square root always be NAN ?
- I don't understand why you need 2 seperate cases depending on the size of adcVal.

Autopilot/Src/MPXV7002DP.cpp
Fberbari: There's more to be done here. Remember from your bootcamp, you needed to put the ADC in constant conversion modeby calling hal_adc_start or something.

Autopilot/Inc/airspeed.hpp
```diff
@@ -84,6 +86,26 @@ class dummyairspeed: public airspeed{
...
+        bool dataNew = false;
+        double airspeed = 0; 
+        int sensorStatus = 0; //Im not 100% sure how to check for errors in the device :(
```
Fberbari: It's fully analog, so we don't really have any way to do that. Only thing you can do is on adc startup, get the error code from the hal function, if something's wrong, you can set status, otherwise, don't worry about it.

Autopilot/Inc/airspeed.hpp
```diff
@@ -16,6 +16,8 @@
 #ifndef AIRSPEED_HPP
```
Fberbari, line 16: Feel free to ad your name as an author to all the files you modify XD

Fberbari (changes requested):
A great start! Some things need to be fixed, but good work. I also really like the way you write coe, it's pretty clean to look at.

Why is Sahil the author of this commit ? lmao

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return 5 * (float)adcVal/(ADC_RESOLUTION-1) - 2.5;
+}
+
+float MPXV7002DP::getAirspeed() {
+    float diffPressure = getDiffPressure();
+    if(diffPressure < 0) {
+        sensorStatus = 1;
+        return 0;
+    } else {
+        sensorStatus - 0;
```
Fberbari: is this meant to be an = sign ?

Autopilot/Inc/airspeed.hpp
```diff
@@ -54,16 +50,11 @@ class airspeed {
          *  ensure that data acquired makes sense, has been
          *  gathered recently within reason (past 10s?)
          * */
-        virtual void GetResult(airspeedData_t &Data) = 0; 
+        virtual void GetResult(airspeedData_t *Data) = 0; 
 };
 
 class dummyairspeed: public airspeed{
```
Fberbari: Go ahead and delete the entire dummy airspeed class, its only reason for existing was to make other shit work while the real class was developed

Fberbari (commented):
Yep, aside from a typo, looks really good. 

Reason the CI fails is because of the whole singleton thing. attitudeStateClasses.hpp and attitudeStateClasses.cpp were built before we decided on singletons and a couple changes need to be made there to accommodate. Look at what was done with the IMU driver in both those files, you'll need to do the same thing with your driver. Ping me if it takes you more than 10 mins.

Fberbari:
Also, can you PR things from your account pls ? We'll fix whatever issue you've got if you can't figure it out, but I don't want you to always be pushing code under Sahil's name...

Autopilot/Src/MPXV7002DP.cpp
SnackkOverflowError: yes, thanks for catching that lol

SnackkOverflowError:
I'll look through the IMU PR and add those changes in a few hours, but just looking through it, I dont think it will be a very complicated fix.

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+void MPXV7002DP::GetResult(airspeedData_t *Data) {
+        Data->airspeed = airspeed;
+        Data->isDataNew = dataNew;
+        dataNew = false;
+        Data->sensorStatus = sensorStatus;
+        Data->utcTime = timeOfLastResult;
+}
+
+void MPXV7002DP::Begin_Measuring() {
```
Fberbari: The Hal_adc_getValue function litterally just reads a register. I.E. It's completely non blocking.
image
So there's no need to give this sensor a thread to begin the measurement process. You should give begin_measuring a blank implementation and explain that it doesn't need to be called in the header, then shift all the code in begin_measuring to get_result. Bonus is that data will be guaranteed to be new on every call.

PR #61, opened 2020-11-18
Waypoint manager

Pop0097:
PR contains the code for the waypoint manager. I have tried to comment the code as much as possible because the math can get confusing at times. To understand what's going on, you can refer to the docs [here](https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1169883137).

You can also refer to PicPilot documentation if need be. The math is very similar because... yk... does math really change?


Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+    uint32_t timeOfData;                // The time that the data in this structure was collected
+} _WaypointManager_Data;
+
+class WaypointManager {
+    public: 
+        /*
+        * Constructor for this class
+        * 
+        * @param _WaypointData * initialWaypoints -> These waypoints will be used to initialize the waypointBuffer array
+        * @param int numWaypoints -> Number of waypoints being initialized in the waypointBuffer array
```
Fberbari: you know what, maybe your param's aren't colorful because you're missing the brackets. 

@param[in] for a variable the module intakes
@param[out] for a variable passed by reference that your module stores return data in.

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+        * @param _WaypointData* currentLocation -> Home base. 
+        */
+        WaypointManager(_WaypointData * initialWaypoints, int numWaypoints, _WaypointData* currentLocation);
+        
+        /*
+        * Updates the _WaypointManager_Data structure with new values
+        * 
+        * @param _Gps_Data currentPosition -> contains the current coordinates, altitude, and heading
+        */
+        void get_next_directions(_Gps_Data currentPosition);
```
Fberbari: So is the idea that every time someone wants to use your module, they at least have to call 

get_next_direction and then GetResult ?

Why are these 2 separate functions ? Could they just be put together into 1 function that takes in input and produces output ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+        * Returns the waypointBuffer array if requested
+        */
+        _WaypointData* get_waypoint_buffer(); //Returns the waypointBuffer array
+
+        /*
+        * Updates an instance of the _WaypointManager_Data structure using the newly calculated values
+        * 
+        * @param _WaypointManager_Data &Data -> Memory address for a structure that holds the data for the state machine
+        */
+        _WaypointManager_Data GetResult(_WaypointManager_Data &Data);
```
Fberbari: Is this function meant to return void ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+        /*
+        * Adds, inserts, updates, or deletes a single waypoint in the waypointBuffer array
+        * 
+        * @param _WaypointData* waypoint -> In the instance that we need to update, insert, or append a new waypoint, this will be used 
+        * @param int waypointId -> the ID of the waypoint that will be updated or deleted
+        * @param _WaypointBufferUpdateType updateType -> the type of modification to the waypointBuffer array (look above)
+        * @param int previousId -> stores the ID of the waypoint that will come before the inserted waypoint. Set to 0 by default, so does not need to be passed
+        * @param int nextId -> stores the ID of the waypoint that will come after the inserted waypoint. Set to 0 by default, so does not need to be passed
+        */
+        void update_path_nodes(_WaypointData* waypoint, int waypointId, _WaypointBufferUpdateType updateType, int previousId = 0, int nextId = 0);
```
Fberbari: confused about the id system. If I just want to add a new way point to the list, what would I do about waypointId ? Does your module expect the caller to keep a list of the way point history ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+* This data will be used by the PID and coordinated turn engine to determine the commands to be sent to the Attitude Manager.
+*/
+typedef struct {
+    uint16_t desiredHeading;            // Desired heading to stay on path
+    int desiredAltitude;                // Desired altitude at next waypoint
+    long double distanceToNextWaypoint; // Distance to the next waypoint (helps with airspeed PID)
+    _WaypointStatus errorCode;          // Contains error codes
+    bool isDataNew;                     // Notifies PID modules if the data in this structure is new
+    uint32_t timeOfData;                // The time that the data in this structure was collected
+} _WaypointManager_Data;
```
Fberbari: LOL so the similarity in the names between this struct and the one above had me confused for a bit there.

If its not too much trouble, could you make the names clearer ? Maybe have "in" in the top one and "out" in the bottom one or something ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+/*
+* Structure stores information about the waypoints along our path to the destination and back. 
+*/
+typedef struct {
+    struct _WaypointData * next;      // Next waypoint
+    struct _WaypointData * previous;  // Previous waypoint
+    long double latitude;             // Latitude of waypoint
+    long double longitude;            // Longitude of waypoint
+    int altitude;                     // Altitude of waypoint
+    float turnRadius;                 // If turn is commanded, this is the radius of the turn
```
Fberbari: hmmm, why are you asking the user to figure out the turn radius ? My expectation was that your module would figure it yourself.

unless you mean this is the radius of a circular hold path ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+    char type;
+    char waypointId;                  // Id of the waypoint
```
Fberbari: what's this ? How does the user know which ID to use ? How does it differ from the IDs mentioned in updatePathNodes() ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+    char arrayIndex;                  // Position of the waypoint within the waypointBuffer (array)
```
Fberbari: Do you expect the user to know something about the array index ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+//Used to specify the modification type when updating the waypointBuffer array
+enum _WaypointBufferUpdateType {ADD_WAYPOINT = 0, UPDATE_WAYPOINT, INSERT_WAYPOINT, DELETE_WAYPOINT};
+
+
+/*
+* Structure stores information about the waypoints along our path to the destination and back. 
+*/
+typedef struct {
+    struct _WaypointData * next;      // Next waypoint
```
Fberbari: Does the user need to know what the previous and next waypoints are ? We're expecting the state machine to store its own list of waypoints ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+enum _WaypointBufferStatus {FREE = 0, FULL};
+
+//Used to specify the modification type when updating the waypointBuffer array
+enum _WaypointBufferUpdateType {ADD_WAYPOINT = 0, UPDATE_WAYPOINT, INSERT_WAYPOINT, DELETE_WAYPOINT};
+
+
+/*
+* Structure stores information about the waypoints along our path to the destination and back. 
+*/
+typedef struct {
```
Fberbari: A lot of my comments on the members of this struct aren't valid if this isn't the case, but this is the struct the state machine would use to ad a new waypoint, right ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+    struct _WaypointData * next;      // Next waypoint
+    struct _WaypointData * previous;  // Previous waypoint
+    long double latitude;             // Latitude of waypoint
+    long double longitude;            // Longitude of waypoint
+    int altitude;                     // Altitude of waypoint
+    float turnRadius;                 // If turn is commanded, this is the radius of the turn
+    char type;
```
Fberbari: type ?

Autopilot/PathManager/Inc/waypointManager.hpp
Fberbari: Is there some way for the user to say: "Start circling here" or "continue going through the waypoint list" or "fuck it return home" ?

I guess default is circling if no waypoints are left in the list ? Or return home ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,175 @@
...
+#define LINE_FOLLOWING 1
+
+#define PI 3.14159265359
+#define MAX_PATH_APPROACH_ANGLE PI/2
+
+#define PATH_BUFFER_SIZE 100
+
+//Waypoint Types
+#define DEFAULT_WAYPOINT 0 // Used to navigate flight path
+#define WAYPOINT_UNUSED 1
```
Fberbari: Where can the user make use of these ?

Fberbari (commented):
Very good stuff here, but I'm also very confused LOOOOOL

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: True, I can blend them together!

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: Yeah. My bad.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: Apparently I needed to put /** when starting the comment instead of /*.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: When appending a waypoint, the id does not need to be passed in separately. I will change that.

Here's how the ID system works:

After a waypoint is created, it is stored in the waypointBuffer array. However, as waypoints are inserted, deleted, etc. the index of the waypoints will change during the duration of the flight. 

For example, say when you initialized the waypointBuffer array, you put w1 at waypointBuffer[2] and w2 at waypointBuffer[3]. However because of deletions and insertions, w1 and w2 are now found at waypointBuffer[5] and waypointBuffer[6]. Now, say you want to insert a new waypoint, w3, between w1 and w2. You can't simply list the array index value since the state machine has no idea what it is. 

Instead, if we give each waypoint an ID (first waypoint has an ID of 1, second 2, and the nth n), you can pass in the ID of the waypoint you care about and the waypoint manager will have a method to find the index of the waypoint with that ID. The state machine will need to be aware of the waypoint IDs and the order with which they appear in the waypointBuffer array, but it will not need to know exactly where they are in the array (the waypoint manager will take care of that)

Hope that helps... I'd be happy to elaborate on Thursday if need be

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: The waypoint IDs are used to have a reliable way of finding the waypoint objects in the waypointBuffer array. 

My understanding is the user or state machine will create the waypoint objects and pass them to the waypoint manager to update the waypointBuffer array. As a result, the state machine should be aware of the IDs of the different waypoints, and the order with which the waypoints appear in the array (according to their ID). 

The IDs in the struct and those in the updatePathNodes() methods are the same.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: No, I only expect them to know the ID of the waypoints (for if they need to update the waypointBuffer array)

I will remove this

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: yeah, so if the user/state machine will need to Create the waypoints using this structure and then pass them into the methods of the waypoint manager. 

This will be used to:
- Create the initial array that initializes the waypointBuffer array (constructor)
- Create the waypoint that symbolizes home base (constructor)
- When calling update_path_nodes() to append, insert or update a waypoint. If you want to delete one, just pass in a blank waypoint

We should move this struct to a separate file, but I am putting it here so my code acc compiles lol

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: My idea was this:

calling void get_next_directions() basically just makes the waypoint manager go through the waypoint list as it normally would, but there are some cases where the plane will trigger certain actions:
- If the waypointType parameter of the waypoint structure is set to 2, then the plane will undergo a holding pattern (start circling)
- If there are no more waypoints present, the plane will continue on it's current/last trajectory (we could change this to make the plane start circling) 

I could create methods that can be called to make the plane start circling/return home.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: First off, I will change WAYPOINT_UNUSED to WAYPOINT_USED

These are used in the type parameter of the waypoint structure. When creating the object, if type = 1, the waypoint management will treat it like a regular waypoint. if type = 2, then the plane will circle when  it reaches the waypoint

Now, as the waypoint manager uses up waypoints, it'll set the type parameter to 1 just to signal that it has been used.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: Just tells the waypoint manager if the waypoint is a regular waypoint or if it is a hold waypoint (when reaching, start circling until further instruction)

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: No, it doesn't need to be. When calling the constructor, if the state machine can pass in the waypoints in the correct order, the waypoint manager can link them together. 

In the parameter list of update_path_nodes(), there are two parameters, **previousId** and **nextId**. When inserting (not appending) a new waypoint into the waypointBuffer array, the state machine only needs to initialize the coordinates, waypointID, altitude, and type. The idea is, using the values for the two parameters, the waypoint manager will link the inserted waypoint with the desired previous and next waypoints.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: Yeah, it's the radius of the circular hold path when type=2.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: Updating variable name to `waypointType`

Fberbari:
Ahh I understand. Some comments might need to be left to explain how to use the ID system at some point, but this looks very good! Good explanations as well, thanks.

A very interesting commit message lmao
image
Ver

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+#define LINE_FOLLOWING 1
+
+//Constants
+#define PI 3.14159265
+#define EARTH_RADIUS 6378.137
+#define PATH_BUFFER_SIZE 100
+#define MAX_PATH_APPROACH_ANGLE PI/2
+
+//Basic Mathematical Conversions
+#define deg2rad(angle_in_degrees) ((angle_in_degrees) * PI/180.0)
```
Fberbari: nitpicky, and I know wer'e gonna put these in a common module anyway, but as a general rule, header files shouldn't be filled with anything the caller wouldn't care about, that's what .cpp files are for. Presenting a super clean, intuitive and simple to use interface should always be the target.

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,236 @@
...
+#define WAYPOINT_MANAGER_H
+
+#include <stdlib.h>
+#include <math.h>
+
+#define ORBIT_FOLLOWING 0
+#define LINE_FOLLOWING 1
+
+//Constants
+#define PI 3.14159265
```
Fberbari: M_PI is actually already defined in math.h

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+    numWaypoints = numberOfWaypoints;
+    nextFilledIndex = 0;
+    nextAssignedId = 0;
+    
+    // Initializes the waypointBuffer array
+    for (int i = 0; i < numWaypoints; i++) {
+        // Links waypoints together
+        if (i == 0) { // If first waypoint, link to next one only
+            initialWaypoints[i].next = &initialWaypoints[i+1]; // Passes address so any updates in initialWaypoints[i+1] is reflected in initialWaypoints[i].next
```
Fberbari: in doubly linked list, general protocol is first_node.previous = NULL; and last_node.next = NULL

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,357 @@
...
+    
+    homeBase = _PathData(); //Initializes homebase first so there are no undefined behaviour issues
+    homeBase = *currentLocation;
+
+    numWaypoints = numberOfWaypoints;
+    nextFilledIndex = 0;
+    nextAssignedId = 0;
+    
+    // Initializes the waypointBuffer array
+    for (int i = 0; i < numWaypoints; i++) {
```
Fberbari, line 94: very nice :D

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,357 @@
...
+    waypoint->longitude = longitude;
+    waypoint->altitude = altitude;
+    waypoint->waypointType = waypointType;
+    // Set next and previous waypoints to empty for now
+    waypoint->next = 0; 
+    waypoint->previous = 0;
+    return waypoint;
+}
+
+_PathData* WaypointManager::initialize_waypoint(long double longitude, long double latitude, int altitude, int waypointType, float turnRadius) {
```
Fberbari: These are great! Only trouble I see is that to use them, you have to have already created an object, which would have required the user to input initial way points in the constructor. I'd consider making these static methods. Makes a lot of sense to me, since they're not really part of of the path management engine as much as a helpful tool.

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,357 @@
...
+    xyCoordinates[1] = get_distance(RELATIVE_LATITUDE, RELATIVE_LONGITUDE, latitude, RELATIVE_LONGITUDE); //Calculates latitude (y coordinate) relative to defined origin (RELATIVE_LONGITUDE, RELATIVE_LATITUDE)
+}
+
+float WaypointManager::get_distance(long double lat1, long double lon1, long double lat2, long double lon2) {
+    // Longitude and latitude stored in degrees
+    // This calculation uses the Haversine formula
+    long double change_in_Lat = deg2rad(lat2 - lat1); //Converts change in latitude to radians
+    long double change_in_lon = deg2rad(lon2 - lon1); //Converts change in longitude to radians
+
+    double haversine_ans = sin(change_in_Lat / 2) * sin(change_in_Lat / 2) + cos(deg2rad(lat1)) * cos(deg2rad(lat2)) * sin(change_in_lon / 2) * sin(change_in_lon / 2); // In kilometers
```
Fberbari, line 237: cool :o didn't know this is how that worked.

Fberbari (commented):
Dude this is awesome work. I really like it so far. It could def benefit from unit testing though, I'll DM you about that.

Autopilot/PathManager/Src/waypointManager.cpp
Pop0097: Yeah, I realized that as I was debugging. I will def transition to this soon!

Pop0097: My facepalm when I noticed it sent my brain to mars.

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,242 @@
...
+#define rad2deg(angle_in_radians) ((angle_in_radians) * 180.0/PI)
+
+//Waypoint Types
+#define DEFAULT_WAYPOINT 0 // Used to navigate flight path
+#define WAYPOINT_USED 1 // Used by the waypint manager to signal that a waypoint has been used
+#define HOLD_WAYPOINT 2 // Circle
+
+// Latitude and longitude of starting area... will need to change accordingly
+#define RELATIVE_LATITUDE 43.473004
+#define RELATIVE_LONGITUDE -80.539678
```
sahil-kale: Could these be defined in a constructor of sort?

sahil-kale (commented):
Looks dope!!
Just wondering about the architecture - are methods like head_home and start_circling considered blocking, or if one is called are they handled in terms of first come first served, or a hierarchy (i.e. head home is higher than start_circling?).

Autopilot/PathManager/Inc/waypointManager.hpp
sahil-kale: defined is the wrong word, set is better LOL

Pop0097: Yes, these could be set in the constructor using gps data. 

Regarding your original comment above:
1. If you want the plane to start circling/heading home, call these functions. What they will do is they will trigger a Boolean variable.
2. continue calling get_next_direction() as normal. Because of the Boolean variables being triggered, there will be an if statement that will be triggered that will make the plane not follow the waypointBuffer array and instead circle, or head towards the waypoint representing home base. 
3. When you want the plane to start following the waypointBuffer array, you can call the start_circling/head_home methods. 

In short, the two methods are like setter methods that set a Boolean variable that makes the plane execute the get_next_directions() method differently. 

So no, they are not blocking methods. 

As for hierarchy, I have yet to decide tbh... but I am thinking of making start_circling have a higher hierarchy (not reflected in code yet). 

You can see the if statements in the image below. You will notice that at the end of the if statement the method terminates. Additionally, if both variables are set to false, the method will just execute the code after the If statements (which I have yet to write).

Screen Shot 2020-11-23 at 4 29 04 PM

I hope this helps. We can discuss more on Tuesday if required.

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,612 @@
+/**
+ * Waypoint Manager Methods and Helpers Implementation
+ * Author: Dhruv Rawat
+ * Created: November 2020
+ * Last Updated: November 2020 (Dhruv)
+ */
+
+#include "../Inc/waypointManager.hpp"
```
Fberbari: cmake takes care of sorting out paths. #include "waypointManager.hpp" is fine

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -5,95 +5,189 @@
...
+//Basic Mathematical Conversions
+#define deg2rad(angle_in_degrees) ((angle_in_degrees) * M_PI/180.0)
+#define rad2deg(angle_in_radians) ((angle_in_radians) * 180.0/M_PI)
+
+//Waypoint Types
+#define DEFAULT_WAYPOINT 0 // Used to navigate flight path
+#define WAYPOINT_USED 1 // Used by the waypint manager to signal that a waypoint has been used
+#define HOLD_WAYPOINT 2 // Circle
+
+// Latitude and longitude of starting area... will need to change accordingly
```
Fberbari: Would you rather these be passed in during initialization of your module ? We don't really want to be changing source code every time we fly in a different area.

Fberbari (commented):
Dude... this is so cool. Looks like you've poured your heart and soul in here, which is kinda awesome. I'm glad so much fire burns inside you. 

I haven't looked at things very deeply in this review, will do that at a later time, but your direction is def good.

Autopilot/PathManager/Src/waypointManager.cpp
Pop0097: yeah, you can pass them in. I will just use them to make the code more readable. Because if I write `if (waypointPointer->waypointType == 1){}`, it is a lot less clear than `if (waypointPointer->waypointType == WAYPOINT_USED){}`

If you think it isn't necessary, I can remove them. No biggie

Fberbari: na you misunderstood my comment. I was reffering to lattitude and longitude of starting area. I don't see how you could know those things at compile time.

(Having macros fir waypoint used and stuff is def better than magic numbers)

Pop0097: Oh lol, yes, we should pass it in!

Pop0097:
Issue: Figure out a way for the program to know when we have passed a waypoint and increment the `currentIndex` parameter.

This will need some thought, but I can continue to develop this module further while I think about this.

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,283 @@
...
+ * Define Waypoint Manager Class and Helpers
+ * Author: Dhruv Rawat
+ * Created: November 2020
+ * Last Updated: December 2020 (Dhruv)
+ */
+
+#ifndef WAYPOINT_MANAGER_H
+#define WAYPOINT_MANAGER_H
+
+#include <iostream>
```
Fberbari: no need anymore.

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,283 @@
...
+     *
+     * Steps:
+     *  - The method will clear the waypointBuffer array
+     *  - Since gps coordiantes of the home base are already stored in the homeBase parameter, the plane will just fly towards that direction
+     *  - As the plane flies, the state machine can call update_path_nodes() to update the waypointBuffer array. The state machine should also keep calling get_next_directions() to get desired heading and altitude
+     *  - Once the waypoints are set, user can call this method again to make the plane follow the waypointBuffer array. (when calling this the second time, goingHome will be set to False)
+     * 
+     *  Returns true if goingHome was set to true. Returns false otherwise. Note that if the homeBase parameter is not initialized, this method will return false automatically
+     */
+    _HeadHomeStatus head_home();
```
Fberbari: This function will be a bit confusing to call in the higher level. ie, if someone is trying to understand the state machine, and they see that I called head_home(), it just looks like I'm trying to head home, even though I could just be stopping the previous process of going home.

I recommend just having a bool parameter that dictates whether to head home or not.

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,952 @@
...
+    for(int i = numWaypoints; i < PATH_BUFFER_SIZE; i++) {
+        waypointBuffer[i] = nullptr;
+    }
+
+    errorStatus = WAYPOINT_SUCCESS;
+    return errorStatus;
+}
+
+_PathData* WaypointManager::initialize_waypoint() {
+    _PathData* waypoint = new _PathData; // Create new waypoint in the heap
```
Fberbari, line 123: Any time you do heap allocation, especially on a safety critical embedded system), you need to check that the allocation succeeded. In the setup you have, if the allocation fails, we'd be de-referencing a null pointer. That puts the processor into hard fault.

I'd recommend returning Null from this function if that occurs ( and telling the user that happens in the prototype). That way the state machine can choose to either go into failure mode or do something else.

Our chip is massive, so running out of memory isn't really a concern, but ZeroPilot has the potential to be a massively complex piece of software  in the far future so maybe that won't always be the case.

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,952 @@
...
+    homeBase = currentLocation; 
+    numWaypoints = numberOfWaypoints;
+    nextFilledIndex = 0;
+    
+    #ifndef UNIT_TESTING
+        currentIndex = 0;
+    #endif
+
+    // Initializes the waypointBuffer array
+    for (int i = 0; i < numWaypoints; i++) {
```
Fberbari: A consideration here is buffer overflow. Best to check if the caller tries providing more way points than can fit.

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,952 @@
...
+    if (numWaypoints != 0) {
+        errorStatus = UNDEFINED_FAILURE;
+        return errorStatus;
+    }
+    
+    homeBase = currentLocation; 
+    numWaypoints = numberOfWaypoints;
+    nextFilledIndex = 0;
+    
+    #ifndef UNIT_TESTING
```
Fberbari: Does current Index not get set unless we're unit testing ? This is different in the overloaded method

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,283 @@
...
+    ~WaypointManager();
+
+    /**
+    * Initializes the flight path
+    *
+    * @param[in] _PathData * initialWaypoints -> These waypoints will be used to initialize the waypointBuffer array
+    * @param[in] int numberOfWaypoints -> Number of waypoints being initialized in the waypointBuffer array
+    * @param[in] _PathData* currentLocation -> Home base.
+    */
+    _WaypointStatus initialize_flight_path(_PathData ** initialWaypoints, int numberOfWaypoints, _PathData *currentLocation); // Sets flight path and home base
```
Fberbari: The implementation of these functions are nearly identical (save for setting a new home base). Having 2 of them overloaded like this just increases the complexity of using the module and makes maintaining their implementations harder. 

My suggestion is to leave currentLocation as a default parameter (set default to nullptr or something). If the user doesn't specify it, then consider they're not changing home base. Doing it this way also means you won't have to modify any of your unit tests.

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,952 @@
...
+
+    // Desired heading
+    int calcHeading = round(90 - rad2deg(courseAngle + turnDirection * (M_PI/2 + atan(k_gain[ORBIT_FOLLOW] * (orbitDistance - turnRadius)/turnRadius)))); //Heading in degrees (magnetic)
+    
+    // Normalizes heading (keeps it between 0.0 and 259.9999)
+    while (calcHeading >= 360.0) {
+        calcHeading -= 360.0; // IF THERE IS A WAY TO DO THIS WITHOUT A WHILE LOOP PLS LMK
+    }
+
+    while (calcHeading < 0.0) {
```
Fberbari: better solution than while loops would be great

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -0,0 +1,283 @@
...
+class WaypointManager {
+public:
+
+    /**
+    * Constructor for this class
+    *
+    * @param[in] float relLat -> This is the relative latitude of the point that will be used as (0,0) when converting lat-long coordinates to cartesian coordiantes. 
+    * @param[in] float relLong -> This is the relative longitude of the point that will be used as (0,0) when converting lat-long coordinates to cartesian coordiantes.
+    */
+    WaypointManager(float relLat, float relLong); // Call this to get an instance of the class
```
Fberbari, line 78: I guess these are pretty arbitrary (i didn't understand that last time I reviewed). In an effort to make things simpler for the caller, wanna just make these be the position of home ? (I guess the variable can be initialized at the time initialize_flight_path gets called)

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -0,0 +1,952 @@
...
+        float euclideanWaypointDirection = sqrt(pow(nextWaypointDirection[0] - waypointDirection[0],2) + pow(nextWaypointDirection[1] - waypointDirection[1],2) + pow(nextWaypointDirection[2] - waypointDirection[2],2)) * ((nextWaypointDirection[0] - waypointDirection[0]) < 0?-1:1) * ((nextWaypointDirection[1] - waypointDirection[1]) < 0?-1:1) * ((nextWaypointDirection[2] - waypointDirection[2]) < 0?-1:1);
+
+        // Determines coordinates of the turn center
+        turnCenter[0] = targetCoordinates[0] + (tangentFactor * (nextWaypointDirection[0] - waypointDirection[0])/euclideanWaypointDirection);
+        turnCenter[1] = targetCoordinates[1] + (tangentFactor * (nextWaypointDirection[1] - waypointDirection[1])/euclideanWaypointDirection);
+        turnCenter[2] = targetCoordinates[2] + (tangentFactor * (nextWaypointDirection[2] - waypointDirection[2])/euclideanWaypointDirection);
+
+        float dotProduct = nextWaypointDirection[0] * (position[0] - halfPlane[0]) + nextWaypointDirection[1] * (position[1] - halfPlane[1]) + nextWaypointDirection[2] * (position[2] - halfPlane[2]);
+        
+        if (dotProduct > 0){
```
Fberbari, line 551: Come to think of it, this might be susceptible to noise. Our sensors aren't perfect, so we might find that the dot product grows larger than 0prematurely because of that jitter. Though this is something to think about in the future rather than in this PR.

Fberbari (commented):
Sweet, took a deep dive into orbit following and did a high level scan of everything else. Looks good. The few comments are more about coding etiquette than functionality of this module. So just do as much as you want tonight and we'll merge when you're good to go.

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: :(( but my iostream... how am I to code in c++ without it?

Fberbari (approved):

Autopilot/PathManager/Src/waypointManager.cpp
Pop0097: OH DAMN

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: I like this

Autopilot/PathManager/Inc/waypointManager.hpp
Pop0097: I'm not too sure what you mean by this

PR #64, opened 2020-11-18
Built theAutoSteer module

Fberbari:


Autopilot/PathManager/Inc/CoordinatedTurns.hpp
```diff
@@ -0,0 +1,50 @@
...
+*/
+void CoordinatedTurns_Init(void);
+
+/**
+* Computes the required commands to deliver to attitude manager to achieve the desired heading as fast as possible and without overshoot.
+* Note that this module does not ever deal with pitch or altitude, that is the work of another module.
+* @param[in]    Input              Input struct containing info about current heading, some other sensor info and desired heading.
+* @param[out]   AttManCommands     Output struct containing commands to be passed down directly to attitude manager.
+*/
+void CoordinatedTurns_Compute(CoordinatedTurnInput_t *Input, TurnAttitudeManagerCommands_t *AttManCommands);
```
Pop0097: `TurnAttitudeManagerCommands_t` should be `CoordnatedTurnAttitudeManagerCommands_t`

Autopilot/PathManager/Inc/CoordinatedTurns.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#endif
```
Pop0097: Please add a new line after the #endif

Pop0097 (commented):
Neat!

Autopilot/PathManager/Inc/CoordinatedTurns.hpp
Fberbari: good catch!

Autopilot/PathManager/Inc/CoordinatedTurns.hpp
Fberbari: It's actually there, if there's no newline, github shows a red minus sign  next to the line to indicate it

Pop0097: Oh, my bad! You learn something new everyday :))

Autopilot/PathManager/src/CoordinatedTurns.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <math.h>
+
+/***********************************************************************************************************************
+ * Definitions
+ **********************************************************************************************************************/
+
+#define RUDDER_SCALING_FACTOR 0.8f // should be experimentally determined
+
+#define DEG_TO_RAD(angleInDegrees) ((angleInDegrees) * M_PI / 180.0)
```
Pop0097: I'm defining this exact same macros (function? what is it called) in my module. We should put this in a more global .hpp file when the time comes for integration.

Fberbari: yea, a macro. Good point, we've got common modules, should keep in mind

Fberbari:
Decided to put in the altitude/airspeed stuff with the coordinated turns stuff. They are both super simple atm and although they aren't now, I could see them being coupled in the future, as we improve ZeroPilot, so this made sense.

Thoughts on this decision ? Also, thoughts on the new name ?

Autopilot/PathManager/src/AutoSteer.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "AutoSteer.hpp"
+#include "PID.hpp"
+
+#include <math.h>
+
+/***********************************************************************************************************************
+ * Definitions
+ **********************************************************************************************************************/
+
+#define RUDDER_SCALING_FACTOR 0.8f // should be experimentally determined
```
sahil-kale: If you're in CPP can we use constexpr?

Autopilot/PathManager/src/AutoSteer.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+void AutoSteer_ComputeAltitudeAndAirspeed(AltitudeAirspeedInput_t *Input, AltitudeAirspeedCommands_t *AttManCommands)
+{
+    float pitchAngle = pitchPid.execute(Input->desiredAltitude, Input->currentAltitude);
+
+    AttManCommands->requiredPitch = DEG_TO_RAD(pitchAngle);
+    AttManCommands->requiredAirspeed = CRUISING_SPEED;      // a simple constant to start with. As things get more complex, circumstances will demand variable airspeeds.
+
+}
```
sahil-kale: This part will work in cruising flight, but for landing, this module shouldn't be used as it will cause the engine to rev up when given a commanded pitch angle to maintain airspeed and will never let the plane touch the ground

sahil-kale (approved):

Autopilot/PathManager/src/AutoSteer.cpp
Fberbari: yea good point. When takeoff and landing are considerations, the complexity of this module will need to be increased.

Autopilot/PathManager/src/AutoSteer.cpp
Fberbari: yea, I'll keep that in mind, thanks.

PR #68, opened 2020-12-09
PWM and Safety Controller addition

sahil-kale:
Added in PWM manager and safety controller with a lot of stuff stripped out.

### DIFF
--- a/Safety/Inc/PWM.hpp
+++ b/Safety/Inc/PWM.hpp
@@ -51,6 +51,7 @@ class PWMChannel {
  void setup();
 
  private:
+	//values in us
 	uint32_t min_signal = 1000; //standard for 50hz pwm signal
 	uint32_t max_signal = 2000;
 };
--- a/Safety/Inc/PWM.hpp
+++ b/Safety/Inc/PWM.hpp
@@ -48,6 +48,7 @@ typedef struct PWMGroupSetting {
 class PWMChannel {
  public:
  void set(uint8_t channel, uint8_t percent);
+ void setup();
 
  private:
 	uint32_t min_signal = 1000; //standard for 50hz pwm signal
--- a/Safety/Inc/safety_controller.hpp
+++ b/Safety/Inc/safety_controller.hpp
@@ -1,0 +1,15 @@
+#ifndef SAFETY_CONTROLLER_H
+#define SAFETY_CONTROLLER_H
+#include "stm32f0xx_hal_iwdg.h"
+#include "PPM.hpp"
+
+/*
+*  Call this function after PPM and PWM manager have been inited.
+*/
+void safety_controller_init();
+void safety_run(PWMChannel &pwm, PPMChannel &ppm);
+bool isSafetyManual();
+void setPWMChannel(PWMChannel &pwm, int channel, int percentage);
+int getPPM(PPMChannel &ppm, int channel);
+
+#endif
--- a/Safety/Inc/safety_controller.hpp
+++ b/Safety/Inc/safety_controller.hpp
@@ -1,15 +1,16 @@
 #ifndef SAFETY_CONTROLLER_H
 #define SAFETY_CONTROLLER_H
-#include "stm32f0xx_hal_iwdg.h"
 #include "PPM.hpp"
 
 /*
 *  Call this function after PPM and PWM manager have been inited.
 */
 void safety_controller_init();
+
+/*
+*  This function needs to get called in the main while loop for constant polling
+*/
 void safety_run(PWMChannel &pwm, PPMChannel &ppm);
 bool isSafetyManual();
-void setPWMChannel(PWMChannel &pwm, int channel, int percentage);
-int getPPM(PPMChannel &ppm, int channel);
 
 #endif
--- a/Safety/Src/PWM.cpp
+++ b/Safety/Src/PWM.cpp
@@ -44,7 +44,7 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
 {
     PWMPinConfig currentChannel = PWM_CONFIG[channel-1];
     uint32_t prescaler = (static_cast<TIM_HandleTypeDef *>(currentChannel.timer))->Init.Prescaler;
-	uint32_t us = ((percent * (PWMChannel::max_signal - PWMChannel::min_signal)) / 100 + min_signal);
+	uint32_t us = ((percent * 10 * (PWMChannel::max_signal - PWMChannel::min_signal)) / 100 + min_signal);
 	uint32_t ticks = (us * (48000000/1000000UL)) / (prescaler + 1);
     __HAL_TIM_SET_COMPARE((TIM_HandleTypeDef *) currentChannel.timer, currentChannel.timer_channel, (uint32_t) ticks);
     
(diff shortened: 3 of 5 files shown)


Safety/Inc/safety_controller.hpp
```diff
@@ -0,0 +1,15 @@
...
+/*
+*  Call this function after PPM and PWM manager have been inited.
+*/
+void safety_controller_init();
+void safety_run(PWMChannel &pwm, PPMChannel &ppm);
+bool isSafetyManual();
+void setPWMChannel(PWMChannel &pwm, int channel, int percentage);
+int getPPM(PPMChannel &ppm, int channel);
+
+#endif
```
Pop0097: Please add a blank line here

Pop0097 (commented):
Not really familiar with the task this is supposed to perform, so I tried to look more at the code and see if I could find anything off. Seems good so far.

Safety/Inc/safety_controller.hpp
sahil-kale: Sounds gucci

Safety/Inc/safety_controller.hpp
```diff
@@ -0,0 +1,15 @@
...
+#define SAFETY_CONTROLLER_H
+#include "stm32f0xx_hal_iwdg.h"
+#include "PPM.hpp"
+
+/*
+*  Call this function after PPM and PWM manager have been inited.
+*/
+void safety_controller_init();
+void safety_run(PWMChannel &pwm, PPMChannel &ppm);
+bool isSafetyManual();
```
Fberbari, line 14: Give each of these a nice description so the user knows how to call them.

Edit: I've just noticed a bunch of these are helper functions only used inside the cpp module. Those kinds of functions should be declared static inside the cpp file, rather than in the header file. Gives the user an easier time trying to figure out how to use your module. As a general rule, only expose (and #include) the bare necessities in the header file.

Safety/Src/PWM.cpp
```diff
@@ -31,12 +31,22 @@ static const PWMPinConfig PWM_CONFIG[MAX_CHANNELS] =
...
+    for(int i = 0; i < MAX_CHANNELS; i++)
+    {
+        PWMPinConfig currentChannel = PWM_CONFIG[i];
+        HAL_TIM_PWM_Start(currentChannel.timer,currentChannel.timer_channel);
+    }
+}
+
 void PWMChannel::set(uint8_t channel, uint8_t percent)
 {
     PWMPinConfig currentChannel = PWM_CONFIG[channel-1];
```
Fberbari, line 50: A check that the user requests a valid channel would be great, and could save someone down the road a lot of trouble.

I'd recommend returning a status integer from this function,

Safety/Inc/safety_controller.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef SAFETY_CONTROLLER_H
+#define SAFETY_CONTROLLER_H
+#include "stm32f0xx_hal_iwdg.h"
```
Fberbari: Does this need to be included here ? Or can i just be included in the cpp file. Reducing dependencies is always on my mind. And actually, this is watchdog shit. Do you really need it in this module ?

Safety/Src/PWM.cpp
```diff
@@ -31,12 +31,22 @@ static const PWMPinConfig PWM_CONFIG[MAX_CHANNELS] =
...
     uint32_t prescaler = (static_cast<TIM_HandleTypeDef *>(currentChannel.timer))->Init.Prescaler;
 	uint32_t us = ((percent * (PWMChannel::max_signal - PWMChannel::min_signal)) / 100 + min_signal);
```
Fberbari: the last min_signal, do you mean PWMChannel::min_signal ?

Math checks out here. All integer division, but in this case it works lol. Just note that it's kinda flimsy. All it takes is for someone to move the "/100" next to percent, and suddenly this gives min_signal every time. I won't ask you to change that though. Might be a learning experience for someone 25 years from now. 

Only thing i recommend is one more bracket that surrounds the stuff before "+min_signal", to avoid confusion.

Safety/Inc/PWM.hpp
```diff
@@ -48,6 +48,7 @@ typedef struct PWMGroupSetting {
 class PWMChannel {
  public:
  void set(uint8_t channel, uint8_t percent);
+ void setup();
 
  private:
 	uint32_t min_signal = 1000; //standard for 50hz pwm signal
```
Fberbari, line 55: A comment to specify micro seconds would be nice. I was confused for a bit.

Safety/Src/PWM.cpp
```diff
@@ -31,12 +31,22 @@ static const PWMPinConfig PWM_CONFIG[MAX_CHANNELS] =
...
-	uint32_t ticks = (us * (48000000/1000000UL)) / (prescaler + 1);
+    uint32_t period = (static_cast<TIM_HandleTypeDef *>(currentChannel.timer))->Init.Period;
+	uint32_t ticks = us*period/(PWMChannel::min_signal*20);
```
Fberbari, line 54: Can confirm math checks out here too. Though its def not clearly written and the "20" would be confusing to anyone not all that familiar with servo PWM.

Same comments with the integer division. if things are rearranged, you'd get problems.

Safety/boardfiles/Src/main.cpp
```diff
@@ -77,6 +78,7 @@
 
 /* Private function prototypes -----------------------------------------------*/
 void SystemClock_Config(void);
+void toggleLED();
```
Fberbari, line 81: This file doesn't look like it's meant for review yet, so I'll leave it un-reviewed.

Safety/Src/safety_controller.cpp
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+    
+}
+
+
+/*
+* Sets an individual PWM channel to the desired output.
+*/
+void setPWMChannel(PWMChannel &pwm, int channel, int percentage)
```
Fberbari: noted earlier, but all helper functions should be declared static and their prototypes should be at the top of the cpp file (so that they're file scope).

Fberbari (approved):
Math checks out. 

I've left a bunch of other comments. They don't need to be addressed before our first test flight. But I'd like them to be addressed right after (though maybe that'l be a job for our next new recruit since we're running out of drivers to write.

Ready to merge into safety-beginnings though, +1 from me.

Safety/boardfiles/Src/main.cpp
sahil-kale: Ya main.cpp is for testing rn lol

Safety/Inc/safety_controller.hpp
sahil-kale: Sounds good, some were originally static but they're bought out since interchip needs to call these functions to transmit (re: Nuclear-safety-changes interchip code) into the struct. I'll def add the descriptions though

Safety/Inc/safety_controller.hpp
sahil-kale: Oop will remove, this was from old safety code

PR #69, opened 2020-12-16
SD Card Header File

layeleomehta:


Autopilot/Src/main.cpp
```diff
@@ -96,6 +96,7 @@ int main(void) {
 
 	size_t bytes_read = 100;
 
+
```
sahil-kale: no need for this change lol

Autopilot/Inc/SDCard.hpp
```diff
@@ -0,0 +1,81 @@
...
+        ACMD41_SD_SEND_OP_COND = 41,
+        ACMD42_SET_CLR_CARD_DETECT = 42,
+        ACMD51_SEND_SCR = 51,
+    };
+};
+
+
+
+
+#endif
```
sahil-kale: Remember to add an extra blank line at the end of files :)

Autopilot/Inc/SDCard.hpp
```diff
@@ -0,0 +1,81 @@
...
+		
+		/*
+		Reads from SD Card
+		*/
+		virtual void Read() = 0; 
+		
+		/*
+		Writes to SD Card
+		*/
+		virtual void Program() = 0;
```
sahil-kale: Curious, do you intend to make another class that will inherit these functions? I don't think an individual card requires its own implementation but I could be wrong since I haven't researched SD cards in-depth 😅

sahil-kale (approved):
Looks good, very minor changes and should be okay!! (I'll approve since they're not earth-shattering ahaha)

Autopilot/Inc/SDCard.hpp
```diff
@@ -0,0 +1,81 @@
+/**
+ * SDcard fuctions
+ * Author: Avani Joshi and Laye Mehta
+ */
+
+#ifndef SDCARDHEADER_HPP
+#define SDCARDHEADER_HPP
+
+#include "SPI.hpp"
```
Fberbari: SPI hal drivers should be included in the .cpp file instead of the header file. Reasoning is that if they're in the header file, anyone who includes your module suddenly knows about the hardware. Not only does that potentially mess up Simulation and unit test builds (which are strictly off the hardware) but it also means there is one more dependency, which is one more place where things can go wrong and one more thing someone using your module potentially has to worry about.

The general rule is to keep interfaces as simple and opaque as possible. That means users can very clearly understand how to use it and what it does for them without having to worry about the magic that goes on inside.  

I also sent ya'll a discord message about some issues about using the SPI.hpp header file rather than the ones directly generated by the hal.

Autopilot/Inc/SDCard.hpp
```diff
@@ -0,0 +1,81 @@
...
+		/*
+		Writes to SD Card
+		*/
+		virtual void Program() = 0; 
+		
+		private:
+    /* Commands : Listed below are commands supported
+     * in SPI mode for SD card : Only Mandatory ones
+     */
+    enum cmdSupported {
```
Fberbari: Sweet! looks like you've figured a few things out about sd cards, good stuff.

Autopilot/Inc/SDCard.hpp
Fberbari: Yeah, to Sahil's point, if communicating with all SD cards is the same, then we probably don't need to have the abstract class.

In case you weren't clear about it, half the reason we have abstract classes is to allow for unit testing against modules (but since your driver is an "optional" feature that doesn't live in the core autopilot architecture, I don't see why anyone would unit test against it).

The other half is that for sensors, if we were to use a different airspeed sensor for example, then we could just create a second derived class from the abstract class for that sensor. The benefit to that being that the person using the driver isn't affected by which sensor is actually used; they just happily call the same base class functions. But since it seems all sd cards are the same (though pls correct me if I'm wrong), no need.

Fberbari (commented):
Yeeee good start!

sahil-kale:
@Fberbari can this be merged?

Fberbari:
@sahil-kale na. There's still a lot of design that needs to be done here. There's also no point. In the past, we merged these interfaces so that people could unit test against them and work on higher level stuff. but an sd card driver isn't part of our main code flow, so no one needs the interface until the implementation also exists.

PR #70, opened 2020-12-16
Waypoint Manager - 2

Pop0097:
There were some errors in the code that got past the Cl compilers that resulting in it not building. This PR address those issues.

Here's what I addressed:

- Included the PathManager directory 
- Added the .cpp files to the "GLOB_RECURSE CXX_SOURCES"
- Added the .hpp files to the "GLOB_RECURSE CXX_INC"

- Included <cstdint> in the .hpp file
- Removed "M_PI" and defined "PI" myself (code wasn't building otherwise)
- Removed warning in waypointManager.cpp where I was declaring a variable and not using it

This resolved all warnings and issues on my computer. I'd recommend that someone build on their own computer (preferably not a Mac) to be safe.

Also, here's what I used to test my build:
- ./Tools/build.bash -c -t
- ./Tools/build.bash -c -r
- ./Tools/build.bash

### DIFF
--- a/Autopilot/CMakeLists.txt
+++ b/Autopilot/CMakeLists.txt
@@ -24,6 +24,7 @@ if (${KIND_OF_BUILD} STREQUAL "FOR_TARGET")
     Libraries/CMSIS/Device/ST/STM32F7xx/Include
     Lib
     AttitudeManager
+    PathManager/Inc
   )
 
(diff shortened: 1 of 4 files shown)


Fberbari (approved):
Warnings and errors fixed. Approved!

PR #72, opened 2020-12-19
Simulation

Fberbari:
This commit contains all the infrastructure that allows ZeroPilot to hook into the Simulation framework that lives in the Simulink-Sim repo (that's where all the cool stuff occurs).

From a user's perspective, the added functionality is the -s flag that can be specified when building. That will build and run the Infinion Simulation, controlled by the attitudeManager state machine. Once the sim is run, the build script will proceed to send the results into FlightGear for vsualization.

In terms of the files changed:
- build.bash and Cmakelists have had extra sections added to them to support the simulation build.
- gitignore ignores the "simulationBuild" directory that gets built when the -s option is specified.
- Imu.hpp and airspeed.hpp have had an extra derived class interface added to them. (That makes them use the intercepts inside Simulink-Sim/ZeroPilot-SWHooks)
- The Simulation directory contains the main.cpp file that governs the simulation as well as the files required to send Simulation results into FlightGear
- All other modified files are indicative of using Infinion, which has a conventional tail, rather than Spike's A tail.


Autopilot/AttitudeManager/GetFromPathManager.cpp
```diff
@@ -4,20 +4,28 @@
...
-}
-
 #else
+	(void) Commands;
 
+	PMError_t errorStruct;
+	errorStruct.errorCode = 0;
+
+	return errorStruct;
 #endif
```
sahil-kale: The random #ifdef makes the code a little unreadable imo

Autopilot/Simulation/SendToFlightGear/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+    QUdpSocket *socket = new QUdpSocket();
+    QHostAddress ipaddr = QHostAddress::LocalHost;
+
+    for (int i = 0; i < numPackets; i++)
+    {
+        socket->writeDatagram(QByteArray((char*)packets[i], strlen(packets[i])), ipaddr, FLIGHTGEAR_UDP_PORT);
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(TIME_BETWEEN_PACKETS_MS));
+    }
+}
```
sahil-kale: I'll be honest I have no clue what this file does, some comments would be nice so that if someone in the future sees this they can understand what's going on 😅

Autopilot/AttitudeManager/GetFromPathManager.cpp
Fberbari: yea you're right, it's not all that pretty. What I might do instead is produce a separate cpp file and link that into the Simulation build instead of this one.

Though I think I'll do that once Anthony actually finishes the actual getFromPathManager implementation.

edit: actually nvm, I'll do that now and re-push. The less the simulation stuff appears in our code files, the better.

Autopilot/Simulation/SendToFlightGear/main.cpp
Fberbari: Once the Simulation is run, and the results are produced, the program that this file compiles to packages the data into a format that flightGear understands and sends it off via a series of UDP packets. FLightGear runs on the same machine in my case, so it just sends them off to a udp port that belongs to the localHost ip address.

I think this explanation belongs in confluence docs on the Simulation architecture rather than in comments. (Although I could deff add a comment that links to those docs).

edit: I've decided to move this file into the simulation repo. I think it makes much more sense to have it there than here.

Autopilot/Tools/build.bash
```diff
@@ -89,6 +94,66 @@ if [[ $RUN_UNIT_TESTS == true ]]; then
...
+        -D KIND_OF_BUILD="SIMULATION"\
+        -Wdev\
+        -Wdeprecated\
+        ../
+    cmake --build $BUILD_DIR
+
+
+    AGE_OF_EXE=$(date -r $BUILD_DIR/sim +"%s")
+
+    if [[ $AGE_OF_EXE -gt $TIME_OF_BUILD ]]; then
```
Pop0097, line 126: Kinda curious to know what `-gt` means. 

I assume it checks if `$AGE_OF_EXE` is less than `$TIME_OF_BUILD` before we reset the simulation results.

Autopilot/Simulation/SimDriver/main.cpp
```diff
@@ -0,0 +1,29 @@
+
+#include "attitudeManager.hpp"
+
+#include "SendInstructionsToSafety.hpp"
+
+/***********************************************************************************************************************
+ * Code
+ **********************************************************************************************************************/
+
+#define NUM_SIMULATION_STEPS 1000
```
Pop0097, line 10: By my understanding, this is the number of times that the autopilot will return a packet of data (contains aircraft position, attitude, etc. [defined in WARG Docs]).

If this is the case, approximately how long will a simulation with 1000 steps be?

Autopilot/Test/Src/AttitudeManager/Test_OutputMixing.cpp
```diff
@@ -259,8 +259,8 @@ TEST(AttitudeManager_OutputMixing, CannotReturnOutOfRangeValuesWhenGivenValidInp
 
 	ASSERT_EQ(error.errorCode, 0);
 
-	ASSERT_TRUE( (mixedOutput[L_TAIL_OUT_CHANNEL] <= 100.0f ) && (mixedOutput[L_TAIL_OUT_CHANNEL] >= -100.0f ) );
-	ASSERT_TRUE( (mixedOutput[R_TAIL_OUT_CHANNEL] <= 100.0f ) && (mixedOutput[R_TAIL_OUT_CHANNEL] >= -100.0f ) );
+	ASSERT_TRUE( (mixedOutput[ELEVATOR_OUT_CHANNEL] <= 100.0f ) && (mixedOutput[ELEVATOR_OUT_CHANNEL] >= -100.0f ) );
```
Pop0097: I would suggest using `EXPECT_TRUE` and `EXPECT_EQ`. That way if a test fails, you will know how many of the comparisons failed instead of only seeing the first one that flopped. 

Had I used this for the waypoint manager, it probs would've shortened debugging time by a bit.

Autopilot/AttitudeManager/OutputMixing.cpp
```diff
@@ -25,17 +32,16 @@ static void constrainOutput(float *channelOut);
...
-	#error "Requested tailtype not supported !"
-#endif
-
 	OutputMixing_error_t error;
 	error.errorCode = checkInputValidity(PidOutput);
 
-    PidOutput->yawPercent += PidOutput->rollPercent * ADVERSE_YAW_MIX; // mix roll rate into rudder to counter adverse yaw
-
+#if TAIL_TYPE == INV_V_TAIL
     channelOut[L_TAIL_OUT_CHANNEL] =  (PidOutput->yawPercent * RUDDER_PROPORTION) - (PidOutput->pitchPercent * ELEVATOR_PROPORTION); //Tail Output Left
```
Pop0097, line 39: in `Autopilot/Test/Src/AttitudeManager/Test_OutputMixing.cpp` you only test if `ELEVATOR_OUT_CHANNEL` and `RUDDER_OUT_CHANNEL` are within [-100f, 100f]. (Line 262)

I may be missing something, but if you are using `L_TAIL_OUT_CHANNEL` and `R_TAIL_OUT_CHANNEL` when `TAIL_TYPE == INV_V_TAIL`, would we want to also test to see if those values are within [-100f, 100f] as well?

Autopilot/AttitudeManager/OutputMixing.cpp
```diff
@@ -4,10 +4,17 @@
...
 
-#define ADVERSE_YAW_MIX 0.5f // This value was taken from picPilot, it might need tuning.
-
-#define RUDDER_PROPORTION 0.75f // has to do with the angle of the inverse V tail, 0.75 means spike's tail is at 45 degrees.
-#define ELEVATOR_PROPORTION 0.75f
+// tails supported by this module
+#define CONVENTIONAL_TAIL 0
+#define INV_V_TAIL 1
+
+#ifdef SPIKE
```
Pop0097, line 11: I may be wrong, but wouldn't SPIKE need to be defined in the CMakeLists.txt file? If so, I don't think you defined it.

If I am mistaken, where are you defining SPIKE?

Autopilot/CMakeLists.txt
```diff
@@ -194,4 +194,69 @@ elseif(${KIND_OF_BUILD} STREQUAL "UNIT_TESTS")
...
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
+
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+
+  include_directories(
+    ${CMAKE_CURRENT_SOURCE_DIR}/../Common/Inc
+    ${CMAKE_CURRENT_SOURCE_DIR}/Inc/
+    ${CMAKE_CURRENT_SOURCE_DIR}/AttitudeManager
```
Pop0097, line 214: Do we not need to include the PathManager directory when building for simulation?

Autopilot/Inc/IMU.hpp
```diff
@@ -42,6 +37,10 @@ class IMU{
         virtual void GetResult(IMUData_t &Data) = 0; //
 };
 
+/***********************************************************************************************************************
+ * Derived classes
```
Pop0097, line 41: YES!! BREAKING CODE INTO SECTIONS USING COMMENTS. LOVE IT!

Autopilot/Inc/IMU.hpp
```diff
@@ -64,8 +63,22 @@ class ICM20602: public IMU{
...
+#ifdef SIMULATION
+
+// This derived class hooks into the Simulink simulation rather than hardware
+class SimulatedIMU : public IMU
+{
+    public :
+        void Begin_Measuring(){};
+        void GetResult(IMUData_t &Data);
+    private:
+        float getNewestDataPoint(const char * fileName);
```
Pop0097, line 80: If you don't mind, could you please provide some clarification about how we will use this class to simulate collecting data?

Pop0097 (commented):
Looking really great so far!

Autopilot/Tools/build.bash
Fberbari: -gt means "greater than", not sure why bash doesn't just use " > ". yes your'e right. The point is that if the exe is newer than the time the build was started, then it was rebuilt by cmake (cmake doesent rebuild if no source files change) and it's necessary to run the simulation again.

Autopilot/Simulation/SimDriver/main.cpp
Fberbari: Actually, this is the number of times Autopilot sends new actuator commands to the safety chip. That frequency is something we haven't chosen yet (though 50-200 Hz is common). 

In Simulink, I set the value to 50 Hz, which means the model updates considers that there are 20 ms intervals between each "Step". 

So 1000 steps gets us 20 seconds of simulated flight.

Autopilot/Test/Src/AttitudeManager/Test_OutputMixing.cpp
Fberbari: Good point. I'll add a task to the asana. I'm thinking we should go through all test files and make this change at some point.

Autopilot/AttitudeManager/OutputMixing.cpp
Fberbari: Not defined yet. All focus is going into Infinion for the moment. though I'm leaving this switch where it's important so that when we get to deal with Spike in a month or so, it'l be easy. The dream is to eventually be able to build with something like:

./Tools/build.bash -AIRCRAFT "SPIKE" and produce the exe tailored to that particular plane or quad or whatever instead of having to manually change source code every time.

Autopilot/AttitudeManager/OutputMixing.cpp
Fberbari: ya, it's a point. I wasn't too sure what to do about that. Cuz testing both setups would mean having to build all the unit tests multiple times (one for each aircraft). I figured we could just unit test the default setup (which is Infinion). If some day someone wants to put in the infrastructure to cover all the setups, that'd be fine, but truthfully, I don't think it's valuable enough to set it up now. (let me know if you disagree)

Autopilot/CMakeLists.txt
Fberbari: Yes we do. But seeing as there's no State machine yet, I can't call it yet. I'll add it in once it's ready

Autopilot/Inc/IMU.hpp
Fberbari: I'd suggest taking a look at https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/924516398/Simulation for how things work but long story short, the implementation of this class exists in https://github.com/UWARG/Simulink-Sim and every time the state machines call GetResult(), they get the data from the matlab model rather than an actual sensor.

Autopilot/Test/Src/AttitudeManager/Test_OutputMixing.cpp
Pop0097: I can do this. I assume the only files that need to be change are in `Autopilot/Test/Src`

sahil-kale (approved):

PR #73, opened 2020-12-23
Safety beginnings ppm

SnackkOverflowError:
added in ppm driver
added in interchip code (no clue whether this will work, but the code looks promising)


Fberbari:
Merging this without deep review. I'll put it under test right away (at least the ppm/pwm stuff) If it checks out, we can move on to thinking about merging into devel.

PR #74, opened 2020-12-24
Changed all ASSERT_'s to EXPECT_'s

Pop0097:
When I was developing the waypoint manager, the ASSERT_ commands made testing a lot harder since it would stop tests when the first condition failed. This wasn't ideal since in some cases multiple tests failed (and knowing this would have helped me find the source of the problem faster). 

Note: I changed all of the files in `./Autopilot/Test/Src/`. Let me know if I missed anything. VS Code said I got rid of em all.

As a result, I believe it is best that we switch all asserts to expects. 

I ran `./Tools/build.bash -c -t` multiple times and all tests passed each time. 

Also cool resource to learn how to cherry pick commits for PRs: https://poanchen.github.io/blog/2017/11/12/How-to-create-a-GitHub-pull-request-with-a-specific-commits


Fberbari (approved):
straightforward.  Approved.

PR #75, opened 2020-12-30
Actual code that Infinion flew with on  first ZeroPilot flight ever

Fberbari:


Fberbari:
A few files where #if 0'd, I was just having some trouble with getting them to build. Also, looks like a lot of files where changed, but obviously the cubemx generated crap doesn't have to be reviewed.

sahil-kale (approved):
Looks good

PR #78, opened 2021-01-20
Fixing hal shitshow

Fberbari:
In this commit, I cleaned up all the driver classes that were manually written. Now, we reference the cubeMX generated drivers.


sahil-kale (approved):
I didn't check 239 files individually, but it looks right and passes the build so 🤷‍♂️

sahil-kale (changes requested):
nvm build failing

sahil-kale (approved):

PR #80, opened 2021-01-21
Telemetry

JTing86:
telemetry merge 2.0


Fberbari:
Sweet. Ok, first thing's first. Your'e gonna have to get all these files building. What I recommend you do is treat the mavlink library as a cmake library. That means you'll have to stick a cmakelists.txt file inside the mavlink directory and tell the cmakelists.txt file in the autopilot directory about it. You'll def have to do some research but one good example is our HAL drivers. Take a look at Autopilot\boardfiles\Drivers\STM32F7xx_HAL_Driver\Cmakelists.txt and then checkout how the top level cmakelists (in AutoPilot) pulls it in as a library.

Autopilot/TelemetryManager/Telemetry_Manager.h
```diff
@@ -0,0 +1,23 @@
+/**
```
Fberbari, line 1: You have to tell CMake about this. You should be able to figure out how just by taking a look at the cmake file. Same for your .cpp file.

Autopilot/TelemetryManager/Telemetry_Manager.h
```diff
@@ -0,0 +1,23 @@
...
+ * @file Telemetry_Manager.h
```
Fberbari, line 2: This file should have a .hpp extension rather than .h. That makes sure the cpp compiler is called and avoids all sorts of garbage problems if you ever need to use cpp code in here.

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
+/**
+ * @file Telemetry_Manager.c
```
Fberbari, line 2: Gotta tell cmake about this and this needs to have a .cpp file extension, same as your header file.

Autopilot/TelemetryManager/Telemetry_Manager.h
```diff
@@ -0,0 +1,23 @@
...
+    WAIT_FOR_GROUND_DATA,
+    CONNECT_WITH_PATH_MANAGER,
+    REGULAR_REPORT_READY,
+    ERR_REPORT_READY,
+    SEND_ERR_REPORT,
+    SEND_REGULAR_REPORT,
+} TM_STATE_e;
+
+void telemetry_manager(void);
+#endif
```
Fberbari, line 23: There needs to be a new line at the end of every file. This has to do with how unix defines lines, and some compilers follow this closely and may screw up.

Autopilot/TelemetryManager/Telemetry_Manager.h
```diff
@@ -0,0 +1,23 @@
...
+    INIT,
+    WAIT_FOR_GROUND_DATA,
+    CONNECT_WITH_PATH_MANAGER,
+    REGULAR_REPORT_READY,
+    ERR_REPORT_READY,
+    SEND_ERR_REPORT,
+    SEND_REGULAR_REPORT,
+} TM_STATE_e;
+
+void telemetry_manager(void);
```
Fberbari, line 22: We have a standard way of doing state machines. It's not too much better than this, but I'd like all our state machines to follow the same structure. Take a look at AutoPilot/AttitudeManager to see how it's done.

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
...
+            case SEND_ERR_REPORT:
+            {
+                // TODO: send data back to ground
+
+                current_state = INIT;
+                break;
+            }
+
+            default:
+                printf("Invalid telemetry manager state: This should never happen\r\n");
```
Fberbari, line 101: Our micro has no support for printf.

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+#include <stdio.h>
+#include <common/mavlink.h>
+
+#include "telemetry_state_machine.h"
+
+
+
+TM_STATE_e current_state = INIT;
```
Fberbari, line 14: variables like this should be declared static, so that they're not global. But this doesen't matter, cuz you should be doing this in a cpp way anyway.

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
...
+
+// TODO: Add FreeRTOS specifics, return type, waits etc
+void telemetery_manager()
+{
+    while (1)
```
Fberbari, line 19: No while(1)! The freeRtos stuff will live above your state machine and will control how it's called. Having 1 call execute 1 cycle of the state machine is sufficient.

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
...
+            case INIT:
+            {
+                // TODO: Insert init code
+                current_state = CONNECT_WITH_PATH_MANAGER;
+                break;
+            }
+            
+            case WAIT_FOR_GROUND_DATA:
+            {
+                while (! (serial.bytesAvailable > 0))
```
Fberbari, line 32: We will never poll. It is based on interrupts and DMA. If there's no data available at the time you want it, you just have to move on. This is done so that each execution of the state machine takes the same amount of time regardless of anything else. That guarantees that we can keep things synchronous.

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    // TODO: FreeRTOS vTaskDelay for x us/ms/whatever: https://www.freertos.org/a00127.html
+                }
+                // received data - mostly encoded GPS coordinate
+                uint8_t byte = serial.getNextByte();
```
Fberbari, line 37: Are these placeholder functions ?

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
+/**
+ * @file Telemetry_Manager.c
+ * @author Jingting Liu
```
Fberbari, line 3: You need to be able to guarantee stateflow works the way you want. We do this with unit tests. Take a look at how it's done for the attitude manager state machine here:
Autopilot\Test\Src\AttitudeManager\Test_Fsm.cpp

I also have a tutorial for how to set up unit tests here: https://uwarg-docs.atlassian.net/wiki/spaces/TUT/pages/1058078728/test+driven+development

Fberbari (changes requested):
Good start! Of course, there's a few things that need to be fixed, but yeah, piece by piece :)

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
...
+ * @date January 14th, 2021
+ */
+
+#include <stdio.h>
+#include <common/mavlink.h>
+
+#include "telemetry_state_machine.h"
```
Pop0097, line 10: I think this is an old filename for "Telemetry_Manager.hpp".

Autopilot/TelemetryManager/Telemetry_Manager.c
```diff
@@ -0,0 +1,104 @@
+/**
+ * @file Telemetry_Manager.c
+ * @author Jingting Liu
+ * @date January 14th, 2021
+ */
+
+#include <stdio.h>
```
Pop0097, line 7: Since we are using C++, <stdio.h> won't work (use <iostream> instead, though I don't think it's needed since we don't use any terminal output/input streams)

Pop0097 (commented):
Looks like a lot of work went into creating this! Great job! 

Just a few comments from me as Anthony covered the rest.

Autopilot/TelemetryManager/Telemetry_Manager.c
JTing86: nice catch, thanks!

Autopilot/TelemetryManager/Telemetry_Manager.c
JTing86: @Pop0097 Hmm okay I'll remove this include, but I can still write this in C right?

Autopilot/TelemetryManager/Telemetry_Manager.c
JTing86: Woah thanks for the super detailed comments, I'll fix them accordingly!

Autopilot/TelemetryManager/Telemetry_Manager.c
Pop0097: Anthony pointed out that he would like for the State machines to have a consistent structure. Since our other state machines use class objects, we'll need to use C++.

PR #81, opened 2021-01-23
Attitude manager cleanup

Pop0097:
What this PR addresses:

- Created an enum in attitudeManager.hpp that indicates the status of the manager: whether it's completed a cycle, in the middle of a cycle, or if it is in failure mode. 
- attitudeStateClasses.cpp -> makes the _PidOutput.yawPercent get its value from GetFromPathManager.cpp instead of depending on the PID module.
- PIDloopMode - Added error catching in the execute() method (attitudeStateClasses.cpp)
- Made each module call the appropriate C-style init methods (if required) (attitudeStateClasses.hpp)
- Handle case where IMU magnetometer data is NAN by just converting it to 0.0f (fetchSensorMeasurementsMode.cpp)
- Added a new state called "fetchSensorMeasurementsMode" that gets sensor data and stores it in new data structures declared in "AttitudeDatatypes.hpp". Goal of the structures is to ensure only fetchSensorMeasurementsMode.hpp has access to the imu and airspeed sensor drivers. 

Let me know if you want me to change anything.

Note this does not address the issue I noted about sensor error handling and it breaking attitude manager. I will have to think more on that and take a Look at the sensor drivers again.


Autopilot/AttitudeManager/attitudeManager.cpp
```diff
@@ -1,15 +1,27 @@
 #include "attitudeManager.hpp"
 #include "attitudeStateClasses.hpp"
+#include "SendInstructionsToSafety.hpp"
```
Pop0097: I dont know why this is here. Will remove when fixing code

Autopilot/AttitudeManager/GetFromPathManager.cpp
```diff
@@ -10,7 +10,13 @@
...
 PMError_t PM_GetCommands(PMCommands *Commands)
 {
-	(void) Commands;
+	// (void) Commands;
+
+	/*** Just for tests to pass ***/
+	Commands->yaw = 0.0f;
+	Commands->roll = 0;
+	Commands->pitch = 0;
+	Commands->airspeed = 0;
```
Pop0097: Will get rid of these when fixing code after suggestions

Autopilot/AttitudeManager/AttitudeDatatypes.hpp
```diff
@@ -9,11 +9,34 @@ typedef struct
...
 #define L_TAIL_OUT_CHANNEL 0 // Spike has ruddervators
 #define R_TAIL_OUT_CHANNEL 1
 #define AILERON_OUT_CHANNEL 2
 #define THROTTLE_OUT_CHANNEL 3
 
+// New datatypes store the IMU and Airspeed data so SensorFusion and other modules
+// do not need to include "IMU.hpp" and "airspeed.hpp"
+struct IMU_Data_t
+{
+	float magx, magy, magz;
```
Pop0097: Will fix indentations here

Pop0097 (commented):
Just some note to selfs... that I made via a GitHub review... it be like that sometimes...

Pop0097:
The merge conflict is legit one line in attitudeStateClasses.cpp -__-

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    //Check if data is old
+    if(!imudata->isDataNew || !airspeeddata->isDataNew){
+        error.errorCode = 1;
+    }
+
+    // Checks if magnetometer values are not a number (NAN) and converts them to zero if they are (ensures Madgwick does not break)
+    // NOTE TO FUTURE DEVELOPERS: At the time of making, our IMU did not have a magnetometer (so for now we set the values to NAN). 
+    // If your IMU does have one, you can remove this
+    if (isnan(imudata->magx)) {
```
Fberbari: I don't think this belongs here. Requiring 0 values is just a quirk of our chosen sensor fusion method. It could be the case that another sensor fusion method wants it some other way. Best to do this inside the sensor fusion function.

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.cpp
```diff
@@ -0,0 +1,80 @@
+/**
+ * Fetch Sensor Measurements Mode: gets raw sensor data from the sensor drivers
+ * Author: Dhruv Rawat
+ * Last edited: January, 2021
+ */ 
+
+#include "fetchSensorMeasurementsMode.hpp"
+#include <math.h>
+
+#include <iostream>
```
Fberbari: not needed, right ?

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+SensorError_t SensorMeasurements_GetResult(IMU *imusns, airspeed *airspeedsns, IMU_Data_t *imudata, Airspeed_Data_t *airspeeddata) {
+
+    SensorError_t error;
+    error.errorCode = 0;
+
+    // Structures from sensor driver files
+    IMUData_t tempIMUdata;
```
Fberbari, line 24: why create these temp variables rather than passing in imudata and airspeeddata ? Seems unnecessary

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.hpp
```diff
@@ -0,0 +1,28 @@
+/**
+ * Fetch Sensor Measurements Mode: gets raw sensor data from the sensor drivers
```
Fberbari, line 2: I think we don't need an entire module dedicated to doing this. All we're really doing is calling the getData functions from the sensors. It doesen't really hurt anyone if we keep this stuff in a module like this, but yeah, sort of overkill and I'd start getting worried about clutter.

Autopilot/Test/Src/AttitudeManager/Test_Fsm.cpp
```diff
@@ -221,6 +289,32 @@ TEST(AttitudeManagerFSM, TransitionFromPIDToOutputMixing) {
 	/**********************ASSERTS**********************/
 
 	EXPECT_EQ(*(attMng.getCurrentState()), OutputMixingMode::getInstance());
+	EXPECT_EQ(attMng.getStatus(), IN_CYCLE);
+
+}
+
+TEST(AttitudeManagerFSM, IfPIDLoopModeFailsTransitionToFailed) {
```
Fberbari, line 296: Unit testing mastered, it seems :))

Autopilot/Test/Src/AttitudeManager/Test_SensorFusion.cpp
```diff
@@ -1,15 +1,19 @@
 /*
-*
 * A skeleton for testing the sensor fusion algorithm
-*
+* Author: Anthony Berbari, Dhruv Rawat
```
Fberbari: This was actually written by Lucy, she never stuck her name in though. You should replace my name with hers.

Fberbari (changes requested):
Very good. A few minor notes, but this is what I hoped for. I'm especially happy with the quality of the unit tests.

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.cpp
Pop0097: Yeah, accidentally left it there

Autopilot/Test/Src/AttitudeManager/Test_SensorFusion.cpp
Pop0097: Sure thing!

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.cpp
Pop0097: I remember why now.

I wanted to remove the `#include "IMU.hpp"` and `"#include "airspeed.hpp"` from as many files as possible. One way I did this was by creating new structures that store the imu and airspeed data.

Now, the sensor drivers' `GetResults()` methods do not accept the new structure I made, so I needed to create temporary IMUData_t and airspeedData_t structures to get the data before initializing the new structure objects I made

Pop0097: If you'd like, I could just use the IMUData_t and airspeedData_t objects from the sensor drivers instead. Only downside is SensorFusion will need `#include "IMU.hpp"` and all

Autopilot/AttitudeManager/fetchSensorMeasurementsMode.hpp
Pop0097: Only reason I made a separate module is so that any calls to the sensor drivers are separated from the rest of the rest of the attitude manager. 

The point you bring up about clutter is fair, though. One solution I can think of is just putting the code in the `SensorMeasurements_GetResult()` function into the `fetchSensorMeasurementsMode::execute()` method. However, if I do this then I might as well ditch the new structs I made since I'll need to `#include "IMU.hpp"` and all in `attitudeStateClasses.hpp`

Fberbari (approved):
I'm happy with this! Good work.

PR #82, opened 2021-01-27
Updated the firmware package of Autopilot

sahil-kale:


Fberbari (approved):
I don't see any obvious problems. Things still look the same in the gui. The true test will come once we upload something to the board of course, but I'm pretty confident that we won't have any problems.

PR #83, opened 2021-01-29
Initial Telemetry Manager 

NixonCTChan:
- created telemetryManager, telemetryStateClasses, telemetryStateManager hpp and cpp files
- modified CMakeLists.txt to include telemetry
- created Test_TelemetryFSM.cpp


Autopilot/TelemetryManager/telemetryManager.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+#include "telemetryStateManager.hpp"
+
+class telemetryState;
+
+class telemetryManager
+{
+    public:
+        telemetryManager();
+        inline telemetryState* getCurrentState() const{return currentState;}
```
sahil-kale, line 13: iirc our compiler doesn't have support for inline - this might be why CI is failing lol

Autopilot/TelemetryManager/telemetryStateClasses.cpp
```diff
@@ -0,0 +1,93 @@
+#include "telemetryStateClasses.hpp"
+#include <time.h>
+
+//definitions
+namespace counter
+{
+    int cycleCounter=0;
+}
+const float CLK_TIME_CONSTANT=1;
```
sahil-kale, line 9: can this be included in the namespace?
Also, why is this a float?

sahil-kale (changes requested):
Looks good! A few changes to address but overall looks really nice
- Comments explaining your logic would be nice as well so that someone 2 years from now understands what you're trying to do

Autopilot/TelemetryManager/telemetryManager.hpp
sahil-kale: Okay nvm looks like we have support for it now... would've saved me a lot of hassle if this was the case 3 months ago 🤔
CI is failing because of an include error on the unit tests.

Autopilot/TelemetryManager/telemetryStateClasses.cpp
sahil-kale, line 9: Put these objects inside the telemetry manager

sahil-kale (changes requested):

Autopilot/TelemetryManager/telemetryManager.hpp
```diff
@@ -0,0 +1,17 @@
...
+        void execute();
+        void setState(telemetryState& newState);
+        bool dataValid;
+        bool dataError;
+    private:
+        telemetryState* currentState;
+};
```
Pop0097: At the end of C and C++ files there needs to be a newline character at the end of the file. This minus sign means the file is missing it. To fix just add a line or two after this closing bracket :))

Autopilot/TelemetryManager/telemetryManager.hpp
```diff
@@ -0,0 +1,17 @@
...
+{
+    public:
+        telemetryManager();
+        inline telemetryState* getCurrentState() const{return currentState;}
+        void execute();
+        void setState(telemetryState& newState);
+        bool dataValid;
+        bool dataError;
+    private:
+        telemetryState* currentState;
```
Pop0097, line 20: Once my PR gets merged, we will need to create an enum that will store the status of the state machine (ex. COMPLETED_CYCLE, IN_CYCLE, FAILURE_MODE).

Autopilot/TelemetryManager/telemetryStateClasses.cpp
Pop0097: Could also use #define

Also if you are going to use a const like this, make it a `static` variable so its scope is to the file only!

Autopilot/TelemetryManager/telemetryStateClasses.hpp
```diff
@@ -0,0 +1,59 @@
...
+        void enter(telemetryManager* telemetryMgr){(void) telemetryMgr;}
+        void execute(telemetryManager* telemetryMgr);
+        void exit(telemetryManager* telemetryMgr){(void) telemetryMgr;}
+        static telemetryState& getInstance();
+
+    private:
+        regularReportMode(){}
+        regularReportMode(const regularReportMode& other);
+        regularReportMode& operator =(const regularReportMode& other);
+};
```
Pop0097, line 143: Same thing here with the newline!

Autopilot/TelemetryManager/telemetryStateClasses.cpp
```diff
@@ -0,0 +1,93 @@
...
+    telemetryMgr->setState(initialMode::getInstance());
+}
+
+telemetryState& errorReportMode::getInstance()
+{
+    static errorReportMode singleton;
+    return singleton;
+}
+
+void regularReportMode:: execute(telemetryManager* telemetryMgr)
```
Pop0097: the space between "::" and "Execute" is messing up gitHub's colouring. It's not a problem, but removing the space in your next commit would make it easier to read through the file

Autopilot/Test/Src/TelemetryManager/Test_TelemetryFSM.cpp
```diff
@@ -0,0 +1,110 @@
...
+    telemetryManager telemetryMng;
+    /***ASSERTS***/
+    ASSERT_EQ(*(telemetryMng.getCurrentState()), initialMode::getInstance());
+}
+
+TEST(TelemetryManagerFSM, InitialTransitionToGround){
+    /***SETUP***/
+    telemetryManager telemetryMng;
+    /***STEPTHROUGH***/
+    telemetryMng.execute();
```
Pop0097, line 39: I think you have to set the state of the telemetryMng before executing the state. 

If you look at the AttitudeManager FSM tests you'll see what I mean

Autopilot/TelemetryManager/telemetryStateManager.hpp
```diff
@@ -0,0 +1,21 @@
...
+        virtual void enter(telemetryManager* telemetryMgr)=0;
+        virtual void execute(telemetryManager* telemetryMgr)=0;
+        virtual void exit(telemetryManager* telemetryMgr)=0;
+
+        bool operator==(const telemetryState& rhs) const{return(this==&rhs);}
+
+        virtual ~telemetryState(){}  
+};
+
+#endif
```
Pop0097: Same here

Autopilot/Test/Src/TelemetryManager/Test_TelemetryFSM.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+TEST(TelemetryManagerFSM, GroundToRegularThroughDataValid){
+    /***SETUP***/
+    telemetryManager telemetryMng;
+    /***DEPENDECIES***/
+    telemetryMng.dataValid=true;
+    telemetryMng.dataError=false;
+    /***STEPTHROUGH***/
+    telemetryMng.execute();
```
Pop0097, line 53: Just curious, but why are you executing the telemetry manager state multiple times?

Pop0097: I don't see the need of executing the telemetry manager twice

Autopilot/Test/Src/TelemetryManager/Test_TelemetryFSM.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+TEST(TelemetryManagerFSM, RegularToGround){
+    /***SETUP***/
+    telemetryManager telemetryMng;
+    /***DEPENDENCIES***/
+    telemetryMng.dataValid=true;
+    telemetryMng.dataError=false;
+    /***STEPTHROUGH***/
+    telemetryMng.execute();//to ground 
+    telemetryMng.execute();//to regular report
```
Pop0097: I see what you are trying to do here. Instead of writing these two lines, just write: 

`telemetryMng.setState(regularReportMode::getInstance());`

That way your code does not depend on the transition from previous parts. That way if the test fails, you are 100% sure it's because of he regualrReportMode state.

Autopilot/TelemetryManager/telemetryStateClasses.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    //form regular report
+    //encode data with Mavlink
+    //send data to ground
+    clock_t time = clock();
+    while((clock()-time)<CLK_TIME_CONSTANT){}
```
Pop0097, line 198: Just a little styling thing for the future, try to put spaces between operators as it makes code much more readable.

Here is the style guide from our old docs: https://www.docs.uwarg.com/computer-vision/coding-conventions/

Autopilot/TelemetryManager/telemetryStateClasses.cpp
Pop0097: instead of declaring a namespace, wouldn't it be better to make a static global variable?

Pop0097 (changes requested):
Great work so far man! This is really good. 

Just some comments and questions from me!

Autopilot/TelemetryManager/telemetryManager.hpp
Pop0097: (It's not appearing here. but go to the "Files changed" section and you'll see it)

Autopilot/TelemetryManager/telemetryStateClasses.hpp
sahil-kale: @Pop0097 Nah defines are a C thing.... we should be moving away from them in CPP files and instead using const since defines don't offer any type protection as it swaps out literals

Autopilot/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ if (${KIND_OF_BUILD} STREQUAL "FOR_TARGET")
     boardfiles/Drivers/CMSIS/Device/ST/STM32F7xx/Include
     AttitudeManager
     PathManager/Inc
+    TelemetryManager
```
Fberbari, line 32: Good, so you've told Cmake to find your header files. Now you just need to tell it to compile your cpp files for the target build. That's done a fe lines below in the statement that starts with "file".

Autopilot/TelemetryManager/telemetryManager.hpp
Fberbari: Yeah. Good point. @NixonCTChan You should do this in this PR. As Dhruv to explain the reasoning behind why its done and how he did it.

Autopilot/TelemetryManager/telemetryStateClasses.cpp
```diff
@@ -0,0 +1,93 @@
...
+    return singleton;
+}
+
+void errorReportMode::execute(telemetryManager* telemetryMgr)
+{
+    //form error report
+    //encode data with Mavlink
+    //report error to ground
+    clock_t time = clock();
+    while((clock()-time)<CLK_TIME_CONSTANT){}
```
Fberbari, line 169: Why do you need this ? In general, no polling should ever be done in any of the state machines. They are super time critical and any time spent by a thread polling potentially takes time away from another thread.

Autopilot/TelemetryManager/telemetryStateClasses.hpp
```diff
@@ -0,0 +1,59 @@
+#pragma once
+#include "telemetryStateManager.hpp"
+#include "telemetryManager.hpp"
+
+class initialMode: public telemetryState
```
Fberbari, line 5: Missing a few states here. I'll fix the telemetry state machine diagram, but in short, your'e missing MavlinkDecode, DataDistrubution (pass to path manager), DataGathering (COllect from Path manager), MavlinkEncode, and SendToGround.

^ I know in the diagram it seems like there's 2 seperate MavlinkEncode states and 2 seperate SendToGroundStates, but you only need 1 of each. I'll make this clear when I update that page.

Autopilot/TelemetryManager/telemetryStateClasses.hpp
Fberbari: @sahil-kale did you mean to respond to this elsewhere ? This comment seems to be about the lack of a new line.

Fberbari: @Pop0097 I'm gonna look into to some git hook which adds newlines automatically lol.

Autopilot/TelemetryManager/telemetryStateManager.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#ifndef TELEMETRYSTATEMANAGER_HPP
```
Fberbari, line 1: Don't need both the #pragma once and the ifndef guard. Either one will do, and I don't really care which one you prefer.

Autopilot/Test/Src/TelemetryManager/Test_TelemetryFSM.cpp
```diff
@@ -0,0 +1,110 @@
...
+    telemetryMng.execute(); //to ground
+    for(int index=0; index<4; index++)
+    {
+        telemetryMng.execute();//to regular report
+        telemetryMng.execute();//to ground
+    }
+    //last increase in counter
+    telemetryMng.execute();//to error report
+    /***ASSERT***/
+    ASSERT_EQ(*(telemetryMng.getCurrentState()), errorReportMode::getInstance());
```
Fberbari: Good stuff! Apart from what Dhruv wrote, a couple of notes:

1. We're ditching ASSERT_EQ in favor of EXPECT_EQ. (I need to update the unit test doc). They serve the same purpose, but ASSERT_EQ kills the test as soon as there's a failure while EXPECT_EQ declares error but keps the test going (helps with debuggabillity).
2. The "dependency" section of the unit test is used for setting up fakes and mocks (which you don't need to do in this case). What you've put inside it belongs in the Setup section.

Fberbari (changes requested):
A very solid start, good work!

Autopilot/TelemetryManager/telemetryStateClasses.cpp
NixonCTChan: I talked with Sahil yesterday, I'm gonna move it into the telemetryManager class.

Autopilot/TelemetryManager/telemetryManager.hpp
NixonCTChan: Got it.

Autopilot/Test/Src/TelemetryManager/Test_TelemetryFSM.cpp
NixonCTChan: got it, will change

Autopilot/TelemetryManager/telemetryStateClasses.cpp
NixonCTChan: will do!

Autopilot/CMakeLists.txt
NixonCTChan: got it!

Autopilot/TelemetryManager/telemetryStateClasses.cpp
NixonCTChan: The Telemetry Manager flowchart includes a loop where it required one CLK time to pass before moving onto another cycle. Do we need that?

Fberbari: Old artifact, I'll remove it from the chart. you don't need to worry about anything like timing inside the state machine.

NixonCTChan:
merged a separate PR

PR #84, opened 2021-02-02
Telemetry Manager Fixed Changes

NixonCTChan:
Fixed changes:
- removal of the time check loop
- addition of more specific states and the testing for those states
- created an enum to define status of the FSM
- removal of the namespace
- merged error report state and regular report state into one state (at this state, errors can be detected and transferred to the report state, which will determine if the FSM has to reinitialize)
- created a fatal fail state and tests 
- modified CMAKE file to try and fix include issue
- rewrote tests to set state rather than repeating execute multiple times
- replaced asserts with expects in unit tests
- fixed pragma once and ifndef


Autopilot/TelemetryManager/telemetryStateClasses.cpp
```diff
@@ -0,0 +1,218 @@
+#include "telemetryStateClasses.hpp"
+#include <time.h>
```
Fberbari: no more need

Autopilot/TelemetryManager/telemetryManager.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
Fberbari, line 5: Around here,we sometimes like to write our names down as Authors at the top of the files we write. Feel free to do that if you like!

Fberbari (approved):
I'm happy with this. It'll have to be re touched up as whoever takes up the telemetry work keeps working on it, but you've done your job here. Approved!!

(Get rid of the time.h thing and add your name if you want, but apart from that, you can go ahead and squash-merge!)

PR #87, opened 2021-02-13
Telemetry-Path Manager Framework

Gfcello:
Building files on Path manager and in Telemetry manager to allow communication between. Also made an interface in the Autopilot/Inc to hold the structs for the data being shared.

File format based off the Path to Attitude manager communication files made by Tony.


Autopilot/PathManager/Inc/TelemToPathManager.hpp
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Initiate communication with the path manager.
+ */
+void CommWithPMInit();
+
+/**
+ * Send attitude/airspeed data to the path manager.
+ * @param[in]   data   Pointer to struct with data.
+ */
+void SendPathData(AttitudeData *data);
```
antimonyB: I think you meant to send telemetry data (like TelemToPMData or commandsForPM depending on the name you choose), not attitude data.

Autopilot/Inc/TelemPathInterface.hpp
```diff
@@ -0,0 +1,49 @@
+/**
+ * Provides declarations common to the telemetry and path manager interfaces.
+ * Author: Anthony Bertnyk, Messed with by Gordon Fountain to turn from Attitude Manager to Telemetry
+ */
+
+#ifndef TELEM_PATH_INTERFACE_HPP
+#define TELEM_PATH_INTERFACE_HPP
+
+//Data for path manager to send to telemetry.
+typedef struct DataForTelem{
```
antimonyB: The struct is named DataForTelem but in other places you use TelemData, which doesn't exist. You probably meant to name this struct TelemData.

Autopilot/Inc/TelemPathInterface.hpp
```diff
@@ -0,0 +1,49 @@
...
+	int errorCode; //Code for specified errors
+	float gpsLattitude,gpsLongitude; //Current gps position
+	int curAltitude,curAirspeed; //Airspeed and altitude data
+	float roll,pitch,yaw;	//Current orientation (radians)
+	bool isLanded; //Switch to check if landed (maybe limit switch?)
+} DataForTelem;
+
+//Data for telemetry to send to path manager.
+typedef struct TelemToPMData{
```
antimonyB: Again, the name TelemToPmData is inconsistent with the name CommandsForPM used elsewhere.

Autopilot/PathManager/Src/TelemToPathManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "TelemToPathManager.hpp"
+
+extern "C"
+{
+#include "cmsis_os.h"
+}
+
+void CommWithPMInit()
+{
+    attitudeDataMailQ = osMailCreate(osMailQ(attitudeDataMailQ), NULL);
```
antimonyB: The mail queue name should also be similar to the name of the struct for sending telemetry data to the path manager (TelemToPMData or commandsWithPM).

antimonyB (commented):
Looks right, besides some naming inconsistencies.

Autopilot/Inc/TelemPathInterface.hpp
Gfcello: Changed it to be more clear about where the data comes from and is going to but forgot I had referenced it elsewhere.

Autopilot/PathManager/Src/PathManagerToTelem.cpp
```diff
@@ -0,0 +1,58 @@
...
+ * Implements methods for sending path manager data to telemetry.
+ * Author: Anthony Bertnyk, Messed with for telemetry use by Gordon Fountain
+ */
+
+#include "PathManagerToTelem.hpp"
+
+extern "C"
+{
+#include "cmsis_os.h"
+}
```
sahil-kale: Curious why the extern C guard on an include?

Gfcello: Tbh, not entirely sure. The include would not get changed by the compiler. I am using the framework Tony created, but I do believe that it would be safe to remove and that my modifications to the code have made it moot.

Autopilot/PathManager/Src/PathManagerToTelem.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+void CommWithTelemInit()
+{
+    attitudeDataMailQ = osMailCreate(osMailQ(telemDataMailQ), NULL);
+}
+
+void SendPathData(PathData *data)
```
antimonyB: PathData isn't declared. Should it be DataForTelem?

Autopilot/PathManager/Src/PathManagerToTelem.cpp
```diff
@@ -0,0 +1,58 @@
...
+    dataOut = static_cast<PathData *>(osMailAlloc(telemDataMailQ, osWaitForever));
+    
+    //Fill mail slot with data
+    *dataOut = *data;
+
+    //Post mail slot to mail queue
+    osMailPut(telemDataMailQ, dataOut);
+}
+
+bool GetCommands(CommandsForPM *commands)
```
antimonyB: CommandsForPM should be TelemToPMData?

Autopilot/TelemetryManager/TelemToPathManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    //Post mail slot to mail queue
+    osMailPut(PMcommandsMailQ, commandsOut);
+}
+
+bool GetTelemData(TelemToPMData *data)
+{
+    //Try to get data from mail queue
+    osEvent event;
+    TelemData * dataIn;
```
antimonyB: I think TelemData should be changed to TelemToPMData.

antimonyB (commented):
Some lingering non-renamed variables.

Autopilot/PathManager/Src/PathManagerToTelem.cpp
antimonyB: I don't know much about mixing C/C++ together but since cmsis_os is written in C I thought this might be safer.

Gfcello: It can't hurt, so I will leave it in for now.

sahil-kale: TLDR it's required if you're using cmsis_os.h functions, I was just curious about it lol

So extern C basically gives a C++ function C linkage. An example of this can be found on the PR I made for FreeRTOS, where I needed to make a C interface for a C++ object. Putting it around an include isn't good practice, but it's required when ppl forgot to keep the piece of code below inside their headers. You'll find the piece of code in a few places where you need to make C compatible with C++. I recommend reading this stack overflow link: https://stackoverflow.com/questions/1041866/what-is-the-effect-of-extern-c-in-c

`#ifdef __cplusplus
extern "C" {
#endif

// all of your legacy C code here

#ifdef __cplusplus
}
#endif`

Autopilot/TelemetryManager/TelemToPathManager.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "TelemToPathManager.hpp"
+
+extern "C"
+{
+#include "cmsis_os.h"
+}
+
+void CommWithPMInit()
+{
+    PMcommandsMailQ = osMailCreate(osMailQ(commandsMailQ), NULL);
```
antimonyB: Changing PMcommandsMailQ and commandsMailQ to be the same name would probably reduce confusion when you refer to it later on.

Autopilot/PathManager/Src/PathManagerToTelem.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "PathManagerToTelem.hpp"
+
+extern "C"
+{
+#include "cmsis_os.h"
+}
+
+void CommWithTelemInit()
+{
+    attitudeDataMailQ = osMailCreate(osMailQ(telemDataMailQ), NULL);
```
antimonyB: attitudeDataMailQ should be telemDataMailQ

Autopilot/TelemetryManager/TelemToPathManager.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * @param[in]   commands   Pointer to struct with commands.
+ */
+void SendCommandsForPM(TelemToPMData *commands);
+
+/**
+ * Retrieve data from telemetry.
+ * @param[out]   data   Pointer to struct to recieve data.
+ * @return True if data was recieved.
+ */
+bool GetTelemData(TelemToPMData *data);
```
antimonyB: Should be receiving DataForTelem.

Autopilot/PathManager/Inc/PathManagerToTelem.hpp
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Initiate communication with telemetry.
+ */
+void CommWithTelemInit();
+
+/**
+ * Send path manager data to telemetry.
+ * @param[in]   data   Pointer to struct with data.
+ */
+void SendPathData(TelemToPMData *data);
```
antimonyB: Since PM is sending data to telem this should be DataForTelem.

antimonyB (commented):
Some minor things.

Autopilot/Inc/TelemDriver.hpp
```diff
@@ -0,0 +1,48 @@
+#ifndef TELEM_DRIVER_HPP
+#define TELEM_DRIVER_HPP
+
+
+#include <ctime>
+#include <cstdint>
+
+struct TelemToPMData {
```
Fberbari: What you've got here is what we've called the PIGO and POGI structs (plane in ground out, plane out ground in). They probably shouldn't be defined in here though because a few different people will need to make use of them. Coordinate with Jingting (working on the mavlink encoder/decoder that you'l be using) and Laye/Avani(working on the path manager state machine) to figure out where everyone wants this data.

Autopilot/Inc/TelemDriver.hpp
```diff
@@ -0,0 +1,48 @@
...
+        float gpsLattitude, gpsLongitude; //Commanded gps position
+        float heading, distance //Commanded heading/distance
+        float roll, pitch, yaw;	//Commanded orientation (radians)
+        bool beginLand; //Command to start landing process
+        bool beginTakeoff; //Starting command
+        bool disconnectAutopilot; //Command to switch off autopilot
+    };
+};
+
+class TelemManager { //Pulling data from ground
```
Fberbari: The telemetry manager class will be a state machine, just like the path and attitude manager classes. We should talk some more about what that means for your implementation.

Autopilot/Inc/TelemDriver.hpp
```diff
@@ -0,0 +1,48 @@
...
+public:
+    virtual void GetResult(TelemToPMData* Data) = 0;
+};
+
+class XBEE : public TelemManager {
```
Fberbari: For the Xbee, your'e going to want to treat it like any other sensor. That means it'll need its own header and implementation files. And one of the states in your state machine will need to ask it for data and give it data to send, which it will be doing periodically, as orchastrated by FREERTos. We'll talk mor about this as well

Autopilot/Inc/TelemDriver.hpp
```diff
@@ -0,0 +1,48 @@
...
+private:
+    XBEE();
+    static XBEE* s_Instance;
+    uint32_t readFromXBEE(uint32_t commandsData);
+    void getRawData(TelemToPMData* commandsData);
+    //That's a lot of data to be recieved from ground, not sure if they need seperate implementation, so keeping this here: float* displaygpsLat, float* displaygpsLong, float* displayHeading, float* displayDistance, float* displayRoll, float* displayPitch, float* displayYaw, bool* LandCommand, bool* liftoffCommand, bool* disconnectAutopilotCommand
+    uint32_t getCurrentTime();
+    uint32_t timeOfResult;
+    bool dataIsNew = false;
+    float gpsLattitudeRead, gpsLongitudeRead; //Commanded gps position
```
Fberbari: This stuff will actually be encoded/decoded by the mavlink stuff, which you should talk to Jingting about some more. I'm not sure if it needs to be stored inside the Xbee class like this.

Fberbari (commented):
You've got some good "guts" here, but the main issue is that the design of the whole system doesn't conform with our standards and there are a few disconnects between what you're doing and the pieces that other people are working on that you'l need. We'll talk through it on Thursday and work out how you should go about things/who you should talk to in a bit more detail.

Autopilot/Inc/TelemDriver.hpp
Gfcello: Hi, I thought as much, so I implemented a header with both PIGO and POGI structs called TelemPathManagerInterface I believe that is also in this PR. I put it here as well because of something I read while trying to understand the Singleton style.

Gfcello: The interface file will be in the autopilot/Inc directory too, so everyone can access it.

PR #88, opened 2021-02-16
path manager changes 2021.2.16

layeleomehta:


Autopilot/PathManager/Inc/pathManager.hpp
```diff
@@ -0,0 +1,24 @@
...
+    public:
+        pathManager();
+        inline pathManagerState* getCurrentState() const {return currentState;}
+        void execute();
+        void setState(pathManagerState& newState);
+        _Path_Manager_Cycle_Status getStatus() {return status;}
+    private:
+        pathManagerState* currentState;
+        _Path_Manager_Cycle_Status status;
+};
```
Pop0097: At the end of .c and .cpp files there needs to be a newline character. This "-" sign means that the character is missing. Just adding a line or two at the end of the affected files will solve the issue :))

Autopilot/PathManager/Inc/pathManager.hpp
```diff
@@ -0,0 +1,24 @@
+// #pragma once is used to ensure that you don't include the same header file more than once?
```
Fberbari: yep pretty much

Autopilot/PathManager/Inc/pathManager.hpp
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+// including the pure abstract class that we will define later on
+#include "pathManagerStateManager.hpp"
+
+//this is the pure abstract class... but why are we just writing the name of the class here?? !QUESTION FLAG!
```
Fberbari: I answered this in our DM, are you still confused about it ? It's the whole circular dependency thing

Autopilot/PathManager/Inc/pathManagerStateManager.hpp
```diff
@@ -0,0 +1,26 @@
...
+class pathManager; //WHY HAVE THIS HERE?!?!? !!QUESTION!!
+
+class pathManagerState
+{
+    public:
+        virtual void enter(pathManager* pathMgr) = 0;
+        virtual void execute(pathManager* pathMgr) = 0;
+        virtual void exit(attitudeManager* pathMgr) = 0;
+
+        bool operator==(const pathManagerState& rhs) const {return (this == &rhs);} //WHAT IS THIS?? !!QUESTION!!
```
Pop0097: This is overloading the "==" operator to check if the memory address of our current state is correct. Since we are looking at class objects, the regular "==" would not work since it checks values (not memory address locations). By overriding, we create this custom functionality that is great for testing and stuff!

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,132 @@
...
+        static AltitudeAirspeedCommands_t *GetPitchAndAirspeed(void) {return &_pitchandairspeed;}
+    private:
+        coordinateTurnElevation() {}
+        coordinateTurnElevation(const coordinateTurnElevation& other);
+        coordinateTurnElevation& operator =(const coordinateTurnElevation& other);
+        CoordinatedTurnAttitudeManagerCommands_t _rollandrudder;
+        AltitudeAirspeedCommands_t _pitchandairspeed; 
+};
+
+class sendToSafetyMode : public pathManagerState
```
Fberbari: Path manager never sends things directly to safety. That jobs belongs to attitude manager

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,132 @@
+#pragma once
```
Pop0097, line 1: Putting comments at the top (similar to other files in our repo) will help other members get an idea of the purpose of the file (and also give yourself credit for making it!)

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,132 @@
...
+        void execute(pathManager* pathMgr);
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+    private:
+        sensorFusion() {}
+        sensorFusion(const sensorFusion& other);
+        sensorFusion& operator =(const sensorFusion& other);
+};
+
+class waypointMgmt : public pathManagerState
```
Fberbari: So rather than calling this the waypointMgmt state, let's call this the "cruising" state. Cuz Nixon is going to add a few more states in here that also use the waypoint manager module for landing.

Fberbari (commented):
Good start! Your'e good to start implementing stuff.

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,132 @@
...
+        void execute(pathManager* pathMgr);
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+        static _PathData *GetWaypointData(void) {return &_waypointdata;}
+        static _WaypointManager_Data_Out *GetOutputData(void) {return &_outputdata;}
+    private:
+        waypointMgmt() {}
+        waypointMgmt(const waypointMgmt& other);
+        waypointMgmt& operator =(const waypointMgmt& other);
+        _PathData _waypointdata;
```
Pop0097: Great job including the data structures for the waypoint manager!

There are some more things that will need to be added:

- A private instance of the waypoint manager class 
- A private array of integers to store the waypoint ids in the order that they are to be executed (If you are statically defining the array, a size of 200 should be good. If you want to make it dynamic, starting off with 50 elements should be perfect). Fun fact, there is a constant `PATH_BUFFER_SIZE` that will help here. It is the size of the flight path array.

Pop0097 (changes requested):
Great job so far! Just some comments on my end!

Autopilot/PathManager/Inc/pathManager.hpp
layeleomehta: oops sorry these are comments I forgot to take out

Autopilot/PathManager/Inc/pathManager.hpp
Pop0097: If you look at the files in the "Files Changed" section you'll see what I mean

Autopilot/PathManager/Src/pathManager.cpp
```diff
@@ -0,0 +1,32 @@
+#include "pathManager.hpp"
+#include "pathManagerStateClasses.hpp"
```
NixonCTChan: Build is failing cause pathManagerStateClasses.hpp doesnt exist, I think its supposed to be pathStateClasses.hpp

NixonCTChan (changes requested):

Autopilot/PathManager/Src/pathManager.cpp
layeleomehta: My bad, I changed the file name to pathStateClasses.hpp and forgot to change the include

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,120 @@
...
+        void execute(pathManager* pathMgr);
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+        static _PathData *GetWaypointData(void) {return &_waypointdata;}
+        static _WaypointManager_Data_Out *GetOutputData(void) {return &_outputdata;}
+    private:
+        cruisingState() {}
+        cruisingState(const cruisingState& other);
+        cruisingState& operator =(const cruisingState& other);
+        //WaypointManager(0,0) {} initializing instance of WaypointManager class
```
Pop0097, line 83: We can declare the waypoint manager object here similar to how sensor class instances are declared in attitude manager!

Since the waypoint manager needs reference coordinates when being initialized, we could create a flag that is sent by telemetry when they send in the original flight path. This flag, when turned on, means that we need to initialize the waypointManager object for this state. Along with this flag will be two coordinates which will be our reference coordinates. 

Then, using an if statement in the execute() method, if the flag is on, we initialize the instance of the waypoint manager!

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,120 @@
...
+        int waypointIDArray[PATH_BUFFER_SIZE]; 
+        static _PathData _waypointdata; 
+        static _WaypointManager_Data_Out _outputdata;
```
Pop0097, line 86: Wait, we need an instance of the waypoint input data struct. This is what we will be sending into waypoint manager when we want to get our desired heading and altitude

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,120 @@
...
+        static pathManagerState& getInstance();
+        static _PathData *GetWaypointData(void) {return &_waypointdata;}
+        static _WaypointManager_Data_Out *GetOutputData(void) {return &_outputdata;}
+    private:
+        cruisingState() {}
+        cruisingState(const cruisingState& other);
+        cruisingState& operator =(const cruisingState& other);
+        //WaypointManager(0,0) {} initializing instance of WaypointManager class
+        int waypointIDArray[PATH_BUFFER_SIZE]; 
+        static _PathData _waypointdata;
```
Pop0097, line 85: You don't need any instances of the _PathData struct inside of the cruisingState state. If we need to create them, they should be made in the execute() method when they are called

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,120 @@
...
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+        static _PathData *GetWaypointData(void) {return &_waypointdata;}
+        static _WaypointManager_Data_Out *GetOutputData(void) {return &_outputdata;}
+    private:
+        cruisingState() {}
+        cruisingState(const cruisingState& other);
+        cruisingState& operator =(const cruisingState& other);
+        //WaypointManager(0,0) {} initializing instance of WaypointManager class
+        int waypointIDArray[PATH_BUFFER_SIZE];
```
Pop0097, line 84: This is good. We'll need to think about managing the elements and all! Any changes to this array should be done after the waypoint manager is run since the module will catch any input errors for us

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -0,0 +1,120 @@
...
+};
+
+class cruisingState : public pathManagerState
+{
+    public:
+        void enter(pathManager* pathMgr) {(void) pathMgr;}
+        void execute(pathManager* pathMgr);
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+        static _PathData *GetWaypointData(void) {return &_waypointdata;}
```
Pop0097, line 77: I don't see any case where we would need this method... I would recommend removing it.

Pop0097 (changes requested):
Just some comments on the crusingState class!

NixonCTChan (approved):

Pop0097 (approved):

PR #89, opened 2021-02-17
Ultrasonic sensor driver implementation 

upadhyaydhruv:
I've created a boilerplate headerfile for the ultrasonic driver, taking inspiration from some of the other sensor drivers. I've marked it as draft.

### DIFF
--- a/Autopilot/Inc/ultrasonic.hpp
+++ b/Autopilot/Inc/ultrasonic.hpp
@@ -32,4 +32,21 @@ class ultrasonic {
         virtual void getData(ultrasonicData_t * Data);
 }
 
-#endif
+class MB1013 : public ultrasonic {
+    public:
+        MB1013(const MB1013*) = delete; // Explicitly declaring the copy constructor, deleting the object as soon as it's made so only one copy can exist
+        static MB1013 * GetInstance();
+        void Init();
+        void beginMeasuring();
+        void getData(ultrasonicData_t * Data);
+
+    private:
+        MB1013();
+        static MB1013* staticInstance;
+        float readFromMB1013();
+        float getTime();
+        float getDistance();
+        float timeMeasured = 0;
+        float distance = 0;
+        uint32_t getCurrentTime();
+}
--- a/Autopilot/Inc/ultrasonic.hpp
+++ b/Autopilot/Inc/ultrasonic.hpp
@@ -1,0 +1,13 @@
+/**
+ * Ultrasonic Sensor Functions
+ * Author: Dhruv Upadhyay, February 2021
+ */
+
+ #ifndef ULTRASONIC_HPP
+ #define ULTRASONIC_HPP
+
+ struct ultrasonicData_t {
+    float distance;
+    bool isDataNew; // Holds whether the sensor data has been updated or not
+    
+ }
--- a/Autopilot/Inc/ultrasonic.hpp
+++ b/Autopilot/Inc/ultrasonic.hpp
@@ -6,12 +6,14 @@
  #ifndef ULTRASONIC_HPP
  #define ULTRASONIC_HPP
 
- #include <stdint.h>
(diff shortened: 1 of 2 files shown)


Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,34 @@
+/**
+ * Ultrasonic Sensor Functions
+ * Author: Dhruv Upadhyay, February 2021
+ */
+
+ #ifndef ULTRASONIC_HPP
+ #define ULTRASONIC_HPP
+
+ #include <stdint.h>
```
sahil-kale: Wait do you need stdint.h?

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+typedef struct {
+    float distance;
+    bool isDataNew; // Holds whether the sensor data has been updated or not
+    uint8_t status; // -1 for trig not sent and echo not received, 0 for trig sent, echo not received, and 1 for echo received
```
sahil-kale: Can we make this an enum?

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,34 @@
...
+    float time; // represents the time taken between trig being sent and echo being received
```
sahil-kale: Is there a reason both time and distance exist as well?

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,34 @@
...
+    const int SOUND_SPEED = 343; // In meters per second, used to calculate distance based on time taken to send and receive ping
```
sahil-kale: Is there a reason that the const is inside the struct as opposed to just living in the class?

sahil-kale (changes requested):
Looks good, just a few changes

Autopilot/Inc/ultrasonic.hpp
upadhyaydhruv: I wanted to make an int with fix width, so that's why I included stdint as well. Since I'm using an enum for status though as you suggested, it isn't required anymore so I'll remove it

Autopilot/Inc/ultrasonic.hpp
upadhyaydhruv: Time is an intermediate variable used to find the distance, tbh I should probably remove it from the struct as it's not a useful part of the final data

Autopilot/Inc/ultrasonic.hpp
upadhyaydhruv: No it should be outside of the struct, mb

NixonCTChan:
Looks good to me, nothing to change.

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,52 @@
...
+ * Author: Dhruv Upadhyay, February 2021
+ */
+
+ #ifndef ULTRASONIC_HPP
+ #define ULTRASONIC_HPP
+
+ #include <stdint.h>
+
+
+const int SOUND_SPEED = 343; // In meters per second, used to calculate distance based on time taken to send and receive ping
```
sahil-kale: Can this be made static inside the cpp file?

sahil-kale (changes requested):

Autopilot/Inc/ultrasonic.hpp
upadhyaydhruv: sounds good

upadhyaydhruv:
Here's a link to the datasheet for reference. I'm using pin 3 (analog representation of range) along with an ADC.
https://www.digikey.ca/htmldatasheets/production/2516798/0/0/1/hrlv-maxsonar-ez-series.html

PR #92, opened 2021-02-19
Autolanding Initial Functions

NixonCTChan:
1st Function: changingAltitude
- calculates the desired altitude for the aircraft at a specific horizontal position
- projects the current location of the aircraft to the plane that is created by the rolling distance vector (stopping point to aiming point) and the slope vector (aiming point to intersection point)
- this plane is basically the vertical plane of the landing path
- by projecting to this vertical plane, the projected point will have x and y coordinates that match a specific z coordinate on the slope line, which is the desired altitude

2nd Function: horizontalAdjustment
- this function should calculate the horizontal adjustment needed when crabbing in the landing path
- I realized that this is exactly what the path manager currently does, and the vertical adjustment can be separated and corrected by the above function
- this function might not even be needed

3rd Function: throttleOff
- this function sets the throttle to zero (in kind of a sketch way, any suggestions?)
- because the PM state machine does not have direct access to controlling the throttle, I think sending an extremely low altitude would work
- this function would not affect the slope and landing path of the aircraft, it is just used for flaring and decrabbing

4th Function: alignHeading
- calculates the heading that is determined by the slope path
- I'm not sure if this is even needed, as CV is supposed to determine a landing direction, which is then used to determine the landing path
- that landing direction can be directly sent to the attitude manager
- difference between heading and track must be kept

5th Function: createSlopeWaypoints
- returns a path structure that contains the 3 important points (stopping, aiming, and intersection point)
- these 3 points are used in other calculations, and can be stored within the path manager state machine?
- calculates those 3 points by using approximate stopping distance (set at 10 m right now), approach angle (set at 5 degrees right now), and landing direction


Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,200 @@
+#include "waypointManager.hpp"
+#include "landingManager.hpp"
+#include <cmath>
+
+const int SET_THROTTLE_OFF = -1000;
+const double DISTANCE_OF_LANDING = 10; //in meters
+const int METERS_PER_DEG_LAT = 111320;
+const int ANGLE_OF_LANDING = 5; //in degrees
```
sahil-kale: Can we make these static?

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+double LandingManager::changingAltitude(_WaypointManager_Data_In * input, _PathData * aimingPoint, _PathData * intersectionPoint, _PathData * stoppingPoint)
+{
+    //arrays to hold the important points
+    float aPoint[3]; //aiming point
+    float iPoint[3]; //intersection point
+    float sPoint[3]; //stopping point
+    float cPoint[3]; //current point
+    float projectedPoint[3];
```
sahil-kale: Is there a reason why float arrays are used here instead of a struct?

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,200 @@
...
+{
+    //determine path and location
+    //using follow waypoints algorithm, output the desired horizontal track
+}
+*/
+double LandingManager::throttleOff(void)
+{
+    //return an altitude so that the throttle turns off
+    //I'm not sure if this is a good way to control throttle within the path manager
+    //Any suggestions?
```
sahil-kale: I don't think turning the throttle on and off is the best way to go about this - it means we'll constantly overshoot and the plane will dolphin on the glide path

sahil-kale (commented):
Looks pretty good - lots of lin alg stuff lol

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: yup!

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: Is there any benefit in using a struct? I think in other files vectors were represented in arrays, that's why I used it here.

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: This is only used for flare and decrab, not any other states. Other states will use the control altitude function, which will send a desired altitude to the attitude manager. 

This function is only used to turn the throttle off when we need to during flare and decrab.

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    cPoint[0] = input->longitude;
+    cPoint[1] = input->latitude;
+    cPoint[2] = input->altitude;
+
+    //calculating the vectors
+    float vectorAI[3];
+    float vectorAS[3];
+    float vectorAC[3];
+    float normal[3];
```
layeleomehta: I think the changingAltitude function looks fine, as ur doing the vector operations exactly as u said in the description. I'm just wondering why the 'normal' array is highlighted? I tried googling it but it doesn't seem to be a keyword?

NixonCTChan: Hmm I'm not sure, I gotta look into that. In VS code it doesn't show up as a keyword, idk why its showing up as a keyword in git.

layeleomehta: Yeah I tried it in VScode too, it didn't highlight weirdly for me either

Autopilot/PathManager/Src/landingManager.cpp
Fberbari: better than using a struct, you should make a class (or find a class ? or maybe C++ has some built in vector type?) specifically for vectors. Then you can overload the +,-,*, and whatever other operations you want to do to be vector addition, vector subtraction, dot product, cross product.... Would make things really clear. I'm sure it'll be used by a ton of other modules in ZeroPilot in time.

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+    vectorAI[0] = iPoint[0] - aPoint[0];
+    vectorAI[1] = iPoint[1] - aPoint[1];
+    vectorAI[2] = iPoint[2] - aPoint[2];
+
+    //aiming point to stopping point
+    vectorAS[0] = sPoint[0] - aPoint[0];
+    vectorAS[1] = sPoint[1] - aPoint[1];
+    vectorAS[2] = sPoint[2] - aPoint[2];
+
+    //aiming point to current point
```
Fberbari: comments like these can be gotten rid of if more thought is put into your variable names.

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+    //normal of the plane created by aiming, stopping, and intersection point
+    normal[0] = vectorAI[1]*vectorAS[2] - vectorAI[2]*vectorAS[1];
+    normal[1] = vectorAI[2]*vectorAS[0] - vectorAI[0]*vectorAS[2]; //flipped negative
+    normal[2] = vectorAI[0]*vectorAS[1] - vectorAI[1]*vectorAS[0];
+
+    //projecting vectorAC to the normal
+    //equation is the dot product of normal and vectorAC, divided by the norm of the normal squared, multiplied by the normal vector
+    float projectionACToNormal[3];
+
+    projectionACToNormal[0] = (normal[0]*vectorAC[0] + normal[1]*vectorAC[1] + normal[2]*vectorAC[2]) / (pow(normal[0],2) + pow(normal[1],2)+ pow(normal[2],2)) * normal[0];
```
Fberbari: https://en.wikipedia.org/wiki/Fast_inverse_square_root

inverse square roots are super expensive computationally. Not a huge deal since we've got a massive processor, but check this out. It's used in MadwickAHRS.c as well

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+    vectorAC[2] = cPoint[2] - aPoint[2];
+    
+    //normal of the plane created by aiming, stopping, and intersection point
+    normal[0] = vectorAI[1]*vectorAS[2] - vectorAI[2]*vectorAS[1];
+    normal[1] = vectorAI[2]*vectorAS[0] - vectorAI[0]*vectorAS[2]; //flipped negative
+    normal[2] = vectorAI[0]*vectorAS[1] - vectorAI[1]*vectorAS[0];
+
+    //projecting vectorAC to the normal
+    //equation is the dot product of normal and vectorAC, divided by the norm of the normal squared, multiplied by the normal vector
+    float projectionACToNormal[3];
```
Fberbari: If I understand correctly, we're already on the glide slope, so the Aiming, current vector should lie on the vertical aiming, stopping, intersection plane. So the projection of this vector onto the normal of that plane should always return the 0 vector, right ?

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    projectionACToNormal[0] = (normal[0]*vectorAC[0] + normal[1]*vectorAC[1] + normal[2]*vectorAC[2]) / (pow(normal[0],2) + pow(normal[1],2)+ pow(normal[2],2)) * normal[0];
+    projectionACToNormal[1] = (normal[0]*vectorAC[0] + normal[1]*vectorAC[1] + normal[2]*vectorAC[2]) / (pow(normal[0],2) + pow(normal[1],2)+ pow(normal[2],2)) * normal[1];
+    projectionACToNormal[2] = (normal[0]*vectorAC[0] + normal[1]*vectorAC[1] + normal[2]*vectorAC[2]) / (pow(normal[0],2) + pow(normal[1],2)+ pow(normal[2],2)) * normal[2];
+
+    //project point onto plane (cpoint minus the projection)
```
Fberbari, line 38: Projecting a point onto a vector isn't a thing. Looks more like you're getting the coordinates of the point that connects the negative vector to point. What's going on here ?

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+    return altitude;
+}
+/*
+double LandingManager::horizontalAdjustment(_WaypointManager_Data_In input, _PathData * aimingPoint, _PathData * intersectionPoint)
+{
+    //determine path and location
+    //using follow waypoints algorithm, output the desired horizontal track
+}
+*/
+double LandingManager::throttleOff(void)
```
Fberbari, line 48: This is too sketchy. There's much better ways to do this (like adding a field in the exchanged data between path manager and attitude manager that specifically kills the throttle). But we'll talk more about that cuz there's a few other complexities (we suport reverse thrust and we'll need another manager for ground operations). For the moment feel free to "turn off the throttle" by having a comment //throttleOff(); in your code.

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+    iPoint[0] = intersectionPoint->longitude;
+    iPoint[1] = intersectionPoint->latitude;
+
+    //determine heading and direction of landing from the aiming point and the landing point
+    dVector[0] = aPoint[0] - iPoint[0];
+    dVector[1] = aPoint[1] - iPoint[1];
+
+    if(dVector[0] >= 0 && dVector[1] > 0) //quadrant 1
+    {
+        heading = atan(dVector[0] / dVector[1]) * 180 / PI;
```
Fberbari: look in to the atan2 function, which exists in the cmath library and deels with all the sign and quadrant crap by itself.

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+*/
+double LandingManager::throttleOff(void)
+{
+    //return an altitude so that the throttle turns off
+    //I'm not sure if this is a good way to control throttle within the path manager
+    //Any suggestions?
+    return SET_THROTTLE_OFF;
+}
+
+double LandingManager::alignHeading(_PathData * aimingPoint, _PathData * intersectionPoint)
```
Fberbari: Looks like this function doesn't deal with crabbing yet. But yeah, waypoint manager already has the functionality you're trying to get into this built in. And we'll add the heading correction in windy conditions to waypoint manager soon as well. So no need for this function.

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+        {
+            heading = 0; //error, the points have the same coordinates, just head north
+        }
+    }
+
+    //return desired heading
+    return heading;
+}
+
+_LandingPath LandingManager::createSlopeWaypoints(_WaypointManager_Data_In * input, float stoppingLatitude, float stoppingLongitude, float stoppingAltitude, double directionLanding)
```
Fberbari, line 80: looks pretty good.

Fberbari (commented):
A pretty good start :)

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: this equation doesn't use inverse square roots though? its just the dot product divided by the norm squared, multipled by the normal?

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: Thats only if the aircraft is perfectly on the glide slope, which we then wouldn't need adjustment for. In that scenario, the current altitude will be returned, as that is the desired altitude for the current horizontal position.

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: A vector is drawn from the aiming point to the current point. That vector is projected onto the normal, to get a vector that is normal to the vertical plane as well as a magnitude that is the distance from the current point to the plane. The resultant vector is a normal to the vertical plane that is drawn to the current point. Given this vector, we can subtract the current point with it, and get a projected point onto the vertical plane.

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: got it!

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: ok!

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: got it!

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: noted!

Autopilot/PathManager/Src/landingManager.cpp
Fberbari: Ah, I see

Autopilot/PathManager/Src/landingManager.cpp
Fberbari: my mistake, thought you were normalizing it. I'll double check your math deeper once you get the vector class in.

Autopilot/PathManager/Src/landingManager.cpp
Fberbari: Ahhh, understood. You're right about calling it projection.

Fberbari (approved):
Good stuff. Merge-able once Travis says OK.

PR #94, opened 2021-02-25
Vectors3D Class and Tests

NixonCTChan:
- created 3D vectors class with various functions 
- created unit tests for those functions
- modified cmakelists.txt to include unit tests


Autopilot/DataStructures/Inc/vectorClass.hpp
```diff
@@ -0,0 +1,48 @@
...
+        double x,y,z;
+
+    public: 
+        Vector3D();
+
+        Vector3D(double inputX, double inputY, double inputZ);
+        
+        ~Vector3D();
+
+        double getX();
```
sahil-kale: add in a const specifier here

sahil-kale (changes requested):
Looks really cool!

Autopilot/Test/Src/DataStructures/Test_Vector_Class.cpp
```diff
@@ -0,0 +1,109 @@
...
+    EXPECT_EQ(vector.getY(), 1);
+    EXPECT_EQ(vector.getZ(), 1);
+}
+
+TEST(VectorClassTest, SettersAndGetters){
+    /***SETUP***/
+    Vector3D vector;
+    vector.setX(5);
+    vector.setY(5);
+    vector.setZ(5);
```
Pop0097: ngl, for tests like this it's better to set each one of them to different values (ex. 1, 2, 3) so you are 100% sure the y value was not set to the x value and stuff. Just makes tests more secure

Autopilot/Test/Src/DataStructures/Test_Vector_Class.cpp
```diff
@@ -0,0 +1,109 @@
...
+    Vector3D vector;
+    /***ASSERTS***/
+    EXPECT_EQ(vector.getX(), 0);
+    EXPECT_EQ(vector.getY(), 0);
+    EXPECT_EQ(vector.getZ(), 0);
+}
+
+TEST(VectorClassTest, DefinedConstructor){
+    /***SETUP***/
+    Vector3D vector(1,1,1);
```
Pop0097: same here. Make each of the coordinates diff values so we know the coordinates were initialized with the correct parameter

Autopilot/DataStructures/Src/vectorClass.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+    z = input;
+}
+
+Vector3D& Vector3D::operator=(const Vector3D &vector2)
+{
+    (*this).x = vector2.x;
+    (*this).y = vector2.y;
+    (*this).z = vector2.z;
+    return *this;
```
Pop0097, line 31: This function returns the address to the Vector3D object, yet we are returning `*this`. I think my knowledge of pointers is breaking down cause I'm pretty confused. Mind explaining what's going on here?

Autopilot/DataStructures/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+project(DataStructures)
+
+file(GLOB_RECURSE DATASTRUCTURES_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/Src/*.c")
+file(GLOB_RECURSE DATASTRUCTURES_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/Inc/*.h")
+
+add_library(${PROJECT_NAME} STATIC ${DATASTRUCTURES_SOURCES} ${DATASTRUCTURES_HEADERS})
+
+target_include_directories(${PROJECT_NAME} PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/Src
+)
```
Pop0097: Add newline

Autopilot/DataStructures/Src/vectorClass.cpp
```diff
@@ -0,0 +1,108 @@
...
+    returnVector.y = (*this).y*coefficient;
+    returnVector.z = (*this).z*coefficient;
+    return returnVector;
+}
+
+Vector3D Vector3D::crossProduct(const Vector3D &vector2)
+{
+    Vector3D returnVector;
+    returnVector.x = (*this).y * vector2.z - (*this).z * vector2.y;
+    returnVector.y = (*this).z * vector2.x - (*this).x * vector2.z; //flipped negative
```
Pop0097: Good call adding the comment here!

Pop0097 (commented):
Looks neat!

Autopilot/Test/Src/DataStructures/Test_Vector_Class.cpp
Pop0097: Seems stupid I know, but it's good practice and may as well apply it here since it's an easy fix

Autopilot/DataStructures/Inc/vectorClass.hpp
NixonCTChan: gotcha!

Autopilot/Test/Src/DataStructures/Test_Vector_Class.cpp
NixonCTChan: for sure, i'll change that

Autopilot/DataStructures/Src/vectorClass.cpp
NixonCTChan: *this refers to the object which the function is called on. I'll use another method if its slightly confusing.

NixonCTChan: ^wait scratch what I said, it works with and without the reference somehow. I changed it so that it returns the original object, so that we can use it for operator chaining.

Autopilot/CMakeLists.txt
```diff
@@ -113,6 +113,7 @@ elseif(${KIND_OF_BUILD} STREQUAL "UNIT_TESTS")
     ${CMAKE_CURRENT_SOURCE_DIR}/Test/Inc
```
Fberbari, line 114: Your'e gonna have to add the library to the target and simulation builds as well. See how that's done with the FREERTOS library.

Autopilot/DataStructures/Inc/vectorClass.hpp
```diff
@@ -0,0 +1,48 @@
...
+    Author: Nixon Chan
+*/
+
+#ifndef VECTOR_CLASS_HPP
+#define VECTOR_CLASS_HPP
+
+class Vector3D
+{
+    private:
+        double x,y,z;
```
Fberbari, line 12: rather than having 3 getter and 3 setter methods, why not just make these variables public? then people can access vector.x really easily and cleanly.

Autopilot/DataStructures/Src/vectorClass.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+double Vector3D::dotProduct(const Vector3D &vector2)
+{
+    return ((*this).x * vector2.x + (*this).y * vector2.y + (*this).z * vector2.z);
+}
+
+double Vector3D::norm()
+{
+    return sqrt(pow((*this).x,2) + pow((*this).y,2) + pow((*this).z,2));
```
Fberbari: "pow" is actually a really slow function (https://baptiste-wicht.com/posts/2017/09/cpp11-performance-tip-when-to-use-std-pow.html). It's really only good for raising things to fractional exponents. But in this case, you can save a fair amount of time by replace pow(*this).x,2) with (this->x * this ->x).

Autopilot/DataStructures/Src/vectorClass.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+void Vector3D::setZ(double input)
+{
+    z = input;
+}
+
+Vector3D Vector3D::operator=(const Vector3D &vector2)
+{
+    (*this).x = vector2.x;
```
Fberbari: The good people who make c++ created arrow notation specifically to avoid the need to worry about brackets around pointer de referencing when accessing members.

(*this).x    is equivalent to      this->x

Autopilot/DataStructures/Src/vectorClass.cpp
```diff
@@ -0,0 +1,108 @@
+/* 
+    Author: Nixon Chan
+*/
+
+#include <math.h>
```
Fberbari: math.h is deprecated in C++, should use cmath. You'll have to call std::sqrt() once you do.

Autopilot/Test/Src/DataStructures/Test_Vector_Class.cpp
```diff
@@ -0,0 +1,109 @@
+#include <gtest/gtest.h>
+#include "math.h"
```
Fberbari: same thing about cmath.

tbh, I'm surprised cmake found this. "" is used for user libraries while <> is used for system libraries.

Fberbari (commented):
I had nothing but nitty gritty comments. Good stuff!

Autopilot/CMakeLists.txt
NixonCTChan: I'm not sure if I'm doing this correctly, but I added the datastructures folder to the Libraries section.

Autopilot/DataStructures/Inc/vectorClass.hpp
NixonCTChan: Fixed!

Autopilot/DataStructures/Src/vectorClass.cpp
NixonCTChan: Fixed!

Autopilot/DataStructures/Src/vectorClass.cpp
NixonCTChan: Fixed!

Autopilot/Test/Src/DataStructures/Test_Vector_Class.cpp
NixonCTChan: Fixed!

Autopilot/CMakeLists.txt
Fberbari: that should be correct. The way to confirm is to just try to instantiate your vector class in some random file and seeing if things compile in the target build.

Fberbari (approved):
Very good. As long as you can confirm that cmake pulls in your stuff in target builds, I'm approving the merge!

sahil-kale (approved):

PR #95, opened 2021-03-01
Vector fixing

NixonCTChan:
- fixing cmakelists.txt can't find library error


Autopilot/PathManager/Src/vectorTesting.cpp
```diff
@@ -0,0 +1,2 @@
+#include "vectorClass.hpp"
```
Fberbari: get rid of this file before merging pls

Fberbari (approved):
Sweet, yeah, just have to remove that testing file and you can go ahead and merge!

PR #96, opened 2021-03-02
FIxed the simulation build

Fberbari:
There was a new file added to the target build that wasn't added to the simulation build causing linker error. That file has now been added.


NixonCTChan (approved):
Looks good!

PR #97, opened 2021-03-07
Autolanding Draft

NixonCTChan:
**NOTE:** This draft is very incomplete, as some parameters and variables are missing and haven't been resolved. The purpose of the draft is to look for suggestions and for any major issues that should be fixed.

General Description:
- Each of the stages in the landing procedure are separated into their own .cpp files. 
- The stages are initialized in the pathStateClasses.hpp file, with public variables that pertain to the functions of each individual stage. 
- Some public pathManager variables are also made (ex. isPackage) for variables that affect the entire landing procedure
- Some baseline constants are made (ex. detect altitudes, speeds). These reference constants can be changed in the future after testing.
- This PR also includes landing functions from a previous PR.

General State Description:
- Most of the states (except the transition state and touchdown state) follow this general structure:
     - detect next stage is reached (if statement for altitude detection)
     - setting horizontal position
     - setting vertical position (whether its throttleOff or desired altitude)
     - setting speed
     - move on to next state (coordinatedTurn or fatalFailure)
- For calculations within all the states, information is retrieved from a single Telemetry_PIGO_t variable (from the getFromTelemetry class) and a single input variable (from the sensorFusion class, not defined). Instructions for desired values (altitude, heading, etc.) are loaded into a single _WaypointManager_Data_Out variable (from the crusingState class, needs speed variable added for speed control)

Other State Descriptions (transition and touchdown):
- Transition overall structure:
     - If landing points have not been made, make them and initialize waypoint management (flight path and waypoints)
     - Get directions
     - If the difference in heading is less than 5 degrees, set to next landing stage

- Touchdown overall structure:
     - Throttle off and maintain heading alignment


Autopilot/PathManager/Inc/landingManager.hpp
```diff
@@ -0,0 +1,64 @@
...
+const static int SET_THROTTLE_OFF = -1000;
+const static double DISTANCE_OF_LANDING = 10; //in meters
+const static int METERS_PER_DEG_LAT = 111320;
+const static int ANGLE_OF_LANDING = 5; //in degrees
+const static int STALL_SPEED_NO_PACKAGE = 20; //in km/h
+const static int STALL_SPEED_WITH_PACKAGE = 30; //in km/h
+const static double PI = 3.14159265358979323846;
+const static int FLARE_ALTITUDE = 2; //in meters
+const static double DECRAB_ALTITUDE = 0.7; //in meters
+const static double TOUCHDOWN_ALTITUDE = 0.05; //in meters
```
sahil-kale: Hmmm, @Fberbari I don't think we should be jamming constants into code but I'm not quite sure how we should implement constants like these.... I was thinking text file but that introduces it's own complications. Thoughts?

Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -0,0 +1,40 @@
+
+#ifndef PATH_DATATYPES_HPP
+#define PATH_DATATYPES_HPP
+
+struct Telemetry_Waypoint_Data_t {
+    long double latitude;
+    long double longitude;
+    int altitude;
+    float turnRadius;
+    int waypointType; // 0 = Path follow, 1 = Orbit, 2 = Hold
```
sahil-kale, line 10: Cab we make this an enum for readability?

Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -0,0 +1,40 @@
...
+};
+
+struct Telemetry_PIGO_t {
+    /* Parameters for the waypoint manager (crusingState) */
+    char numWaypoints;
```
sahil-kale, line 15: curious why the use of the char?

Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -0,0 +1,40 @@
...
+    
+    char waypointModifyFlightPathCommand; // 0 = nothing, 1 = initialize flight path, 2 = append, 3 = insert, 4 = update, 5 = delete, 6 = nuke
```
sahil-kale, line 17: Can we make this into an enum again ahaha

Autopilot/PathManager/Src/landingManager.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+const static int SET_THROTTLE_OFF = -1000;
+const static double DISTANCE_OF_LANDING = 10; //in meters
+const static int METERS_PER_DEG_LAT = 111320;
+const static int ANGLE_OF_LANDING = 5; //in degrees
+const static int STALL_SPEED_NO_PACKAGE = 20; //in km/h
+const static int STALL_SPEED_WITH_PACKAGE = 30; //in km/h
+const static double PI = 3.14159265358979323846;
+const static int FLARE_ALTITUDE = 2; //in meters
+const static double DECRAB_ALTITUDE = 0.7; //in meters
```
sahil-kale: Curious are some of these constants redefined twice?

sahil-kale (changes requested):
Holy crap this looks amazing @NixonCTChan !! Great work so far :)

I've left some minor changes for you to address

Autopilot/PathManager/Inc/pathDatatypes.hpp
NixonCTChan: lol this structure was created by dhruv, no credit to me. I just implemented it so that I had some parameters for the functions. @Pop0097

Autopilot/PathManager/Src/landingManager.cpp
NixonCTChan: oh yeah true, fixed!

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -55,13 +54,41 @@ pathManagerState& getSensorData::getInstance()
...
-    if(isError)
+    if(pathMgr->isError)
     {
         pathMgr -> setState(fatalFailureMode::getInstance());
     }
     else
     {
-        pathMgr -> setState(cruisingState::getInstance());
+        //if the enums for landing state, set to each landing state
+        if(pathMgr->stage == TRANSITION)
```
Fberbari: switch statement is nicer for this sort of thing.

Autopilot/PathManager/Src/landingTransitionStage.cpp
```diff
@@ -0,0 +1,65 @@
+#include "pathStateClasses.hpp"
+
+void landingTransitionStage::execute(pathManager* pathMgr)
```
Fberbari: About this, I  don't want the actual implementations of the states scattered across multiple files. State machine design principles say that makes it pretty hard for someone to try and trace back how things transition. What you should do is have this function (along with the transition logic) inside the stateClasses.cpp file. That being said, feel free to store the guts of it in some other function in some other file and just call that function from the stateClasses file. You'l have to pass in all the variables you need from the other states as arguments.

Autopilot/PathManager/Src/landingTransitionStage.cpp
```diff
@@ -0,0 +1,65 @@
...
+        //set made madelandingPoints to true
+        pathMgr->madeLandingPoints = true;
+    }
+
+    //follow the landing waypoints
+    waypointStatus = landingPath.get_next_directions(getFromTelemetry::telemetryInput, &cruisingState::_outputdata);
+
+    //calculating the difference in heading to detect if finished turning (2 differences in heading possible)
+    differenceInHeading1 = getFromTelemetry::telemetryInput.landingDirection - sensorFusion::input.track;
+    differenceInHeading2 = sensorFusion::input.track - getFromTelemetry::telemetryInput.landingDirection;
```
Fberbari: One of these is just the negative of the other atm. Is that what you want ?

Autopilot/PathManager/Inc/pathDatatypes.hpp
Pop0097: Could create an enum here, but it'll add complexity to CommsWithTelemetry state. Will add it once everything is taking shape to prevent merge conflicts and all

Autopilot/PathManager/Inc/pathDatatypes.hpp
Pop0097: Should be int

Autopilot/PathManager/Src/landingTouchdownStage.cpp
```diff
@@ -0,0 +1,22 @@
+#include "pathStateClasses.hpp"
+
+void landingTouchdownStage::execute(pathManager* pathMgr)
+{
+    //throttleOff()
+    //aligning heading
+    cruisingState::_outputdata.desiredHeading = getFromTelemetry::telemetryInput.landingDirection;
```
Fberbari: why do you need to modify cruising state variables from here ? I'm weary about doing this since it can lead to spaghetti code.

Autopilot/PathManager/Inc/landingManager.hpp
```diff
@@ -0,0 +1,63 @@
+#pragma once
+#include "waypointManager.hpp"
+
+const static double DISTANCE_OF_LANDING = 10; //in meters
```
Fberbari: To Sahil's point, we should thing about storing these somewhere else, not really sustainable to have them scattered al over the place. 

To my roommate's point, it turns out C++ has something called a constexpr, which guarantees compile time evaluation.

Fberbari (commented):
I'm genuinely impressed. I really can't wait to see this in action. Keep up the great work!

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: got it!

Autopilot/PathManager/Src/landingTransitionStage.cpp
NixonCTChan: I'll probably just move everything into the stateClasses.cpp file then. The only long function is the transition state, so it wont add too many lines to the file.

Autopilot/PathManager/Src/landingTransitionStage.cpp
NixonCTChan: The reason I'm doing this is to ensure tracks that are around 360 are accounted for. So for example, the landing direction is 355 and the current track is 10 degrees. I would get a differenceInHeading1 = 345, and the differenceInHeading2 =-345. Then, I add 360 to the directions if they are negative, so differenceInHeading1 = 345 and the differenceInHeading2 = 15. This results in the smaller heading being the difference we want. In this example, that would be 15 degrees rather than 345 degrees in the other direction.

Autopilot/PathManager/Src/landingTouchdownStage.cpp
NixonCTChan: It's so that landing directions would only be loaded into one variable. I think that it would be inefficient for coordinatedTurns to pull landing instructions from multiple variables across different states, therefore loading it into crusingState output would allow coordinatedTurns to pull instructions from one variable.

NixonCTChan:
- landing and takeoff finished!

PR #99, opened 2021-03-19
Wind and attitude control fixes

NixonCTChan:
This draft consists of two separate fixes, the wind fix and the attitude control fix.

**Wind Fix Changes**
- switched heading to track in waypoint manager
- added speed and track entries to the output structure, and current track to the input structure (waypoint manager)
- added trackFinder file (contains getTrack function) (I have the test in another file, not in this PR)

**Attitude Control Changes**
- changed PM_Command entires to include throttle percent, rudder percent, and removed airspeed and yaw
- changed the way airspeed and altitude was being controlled
- changed the way bank angle was being controlled (switched heading to headingTrack)
     - this gives us the option to either control the heading or the track of the plane, depending on what is passed into the parameters

Attitude resturcture diagram:
image

### DIFF
--- a/Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
+++ b/Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
@@ -88,14 +88,14 @@ void PIDloopMode::execute(attitudeManager* attitudeMgr)
     PMCommands *PMInstructions = fetchInstructionsMode::GetPMInstructions();
     SFOutput_t *SFOutput = sensorFusionMode::GetSFOutput();
 
-    // Gets roll, pitch, yaw, and airspeed commands from the path manager module
+    // Gets roll, pitch, rudder, and throttle commands from the path manager module
     PMCommands pathManagerOutput;
     PMError_t pmError = PM_GetCommands(&pathManagerOutput);
 
     _PidOutput.rollPercent = _rollPid.execute(PMInstructions->roll, SFOutput->IMUroll, SFOutput->IMUrollrate);
     _PidOutput.pitchPercent = _pitchPid.execute(PMInstructions->pitch, SFOutput->IMUpitch, SFOutput->IMUpitchrate);
-    _PidOutput.yawPercent = pathManagerOutput.yaw;
-    _PidOutput.throttlePercent = _airspeedPid.execute(PMInstructions->airspeed, SFOutput->Airspeed);
+    _PidOutput.yawPercent = pathManagerOutput.rudderPercent;
+    _PidOutput.throttlePercent = pathManagerOutput.throttlePercent;
 
     if (pmError.errorCode == 0) 
     {
--- a/Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
+++ b/Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
@@ -94,7 +94,7 @@ void PIDloopMode::execute(attitudeManager* attitudeMgr)
 
     _PidOutput.rollPercent = _rollPid.execute(PMInstructions->roll, SFOutput->IMUroll, SFOutput->IMUrollrate);
     _PidOutput.pitchPercent = _pitchPid.execute(PMInstructions->pitch, SFOutput->IMUpitch, SFOutput->IMUpitchrate);
-    _PidOutput.yawPercent = pathManagerOutput.rudderPercent;
+    _PidOutput.rudderPercent = pathManagerOutput.rudderPercent;
     _PidOutput.throttlePercent = pathManagerOutput.throttlePercent;
 
     if (pmError.errorCode == 0) 
--- a/Autopilot/PathManager/Inc/AutoSteer.hpp
+++ b/Autopilot/PathManager/Inc/AutoSteer.hpp
@@ -20,12 +20,9 @@
 
 typedef struct
 {
-    float currentHeading;    // degrees
-    float desiredHeading;    // degrees
-    float currentTrack;
-    float desiredTrack;
+    float currentHeadingTrack;    // degrees
+    float desiredHeadingTrack;    // degrees
     float accY;              // Y axis accelerometer reading from the IMU.
-    bool useHeading;
 }CoordinatedTurnInput_t;
 
 typedef struct
--- a/Autopilot/PathManager/Inc/AutoSteer.hpp
+++ b/Autopilot/PathManager/Inc/AutoSteer.hpp
@@ -20,6 +20,16 @@
 
 typedef struct
 {
+    /*
+    By using currentHeadingTrack, the user can pass in whatever type of parameter they want to control
+    For example, if they wanted to maintain a heading, simply pass in the desired heading and current heading
+    If track needs to be controlled, simply pass tracks into these entries
+
+    Each path manager stage will have its own control needs for its functions. Doing this will allow each
+    stage to control the plane's attitude more precisely.
+    Ex. waypoint manager needs to control track, but landingDecrabStage needs to control heading
+    */
+
     float currentHeadingTrack;    // degrees
     float desiredHeadingTrack;    // degrees
     float accY;              // Y axis accelerometer reading from the IMU.
--- a/Autopilot/PathManager/Inc/AutoSteer.hpp
+++ b/Autopilot/PathManager/Inc/AutoSteer.hpp
@@ -20,16 +20,20 @@
 
 typedef struct
 {
-    float currentTrack;    // degrees
-    float desiredTrack;    // degrees
+    float currentHeading;    // degrees
+    float desiredHeading;    // degrees
+    float currentTrack;
+    float desiredTrack;
     float accY;              // Y axis accelerometer reading from the IMU.
-
+    bool useHeading;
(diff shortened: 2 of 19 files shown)


Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -88,14 +88,14 @@ void PIDloopMode::execute(attitudeManager* attitudeMgr)
...
-    // Gets roll, pitch, yaw, and airspeed commands from the path manager module
+    // Gets roll, pitch, rudder, and throttle commands from the path manager module
     PMCommands pathManagerOutput;
     PMError_t pmError = PM_GetCommands(&pathManagerOutput);
 
     _PidOutput.rollPercent = _rollPid.execute(PMInstructions->roll, SFOutput->IMUroll, SFOutput->IMUrollrate);
     _PidOutput.pitchPercent = _pitchPid.execute(PMInstructions->pitch, SFOutput->IMUpitch, SFOutput->IMUpitchrate);
-    _PidOutput.yawPercent = pathManagerOutput.yaw;
-    _PidOutput.throttlePercent = _airspeedPid.execute(PMInstructions->airspeed, SFOutput->Airspeed);
+    _PidOutput.yawPercent = pathManagerOutput.rudderPercent;
```
Fberbari: Might want to change this to rudderPercent.

Autopilot/PathManager/Inc/AutoSteer.hpp
```diff
@@ -20,16 +20,17 @@
 
 typedef struct
 {
-    float currentHeading;    // degrees
-    float desiredHeading;    // degrees
+    float currentHeadingTrack;    // degrees
```
Fberbari, line 33: This is a bit unclear. A couple comments explaining how these fields are used would go a long way.

Autopilot/PathManager/Inc/trackFinder.hpp
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+double getTrack(double latitudeInitial, double longitudeInitial, double latitudeCurrent, double longitudeCurrent);
```
Fberbari: Even if it seems obvious, your prototype needs to have comments explaining all parameters,the return value, what the function does, and how it should be called. This is meant to go into sensor fusion once that's merged, right ?

Autopilot/PathManager/Src/AutoSteer.cpp
```diff
@@ -62,10 +62,10 @@ void AutoSteer_ComputeCoordinatedTurn(CoordinatedTurnInput_t *Input, Coordinated
 
 void AutoSteer_ComputeAltitudeAndAirspeed(AltitudeAirspeedInput_t *Input, AltitudeAirspeedCommands_t *AttManCommands)
 {
-    float pitchAngle = pitchPid.execute(Input->desiredAltitude, Input->currentAltitude);
+    float pitchAngle = pitchPid.execute(Input->desiredAirspeed, Input->currentAirspeed);
```
Fberbari, line 65: This may be true on approach, but is it also just as true on climb and in cruising flight ? Would you mind looking into how this was done on picPilot, and maybe doing a bit more research ?

I'm worried this may actually be a bit too simplistic. You may also want to look into how other autopilots do it.

Autopilot/PathManager/Src/trackFinder.cpp
```diff
@@ -0,0 +1,18 @@
+#include <math.h>
+#include "trackFinder.hpp"
+
+double getTrack(double latitudeInitial, double longitudeInitial, double latitudeCurrent, double longitudeCurrent)
+{
+    double deg2rad = 3.14159265358979323/180;
```
Fberbari: this should be a constexpr.

Autopilot/PathManager/Src/trackFinder.cpp
```diff
@@ -0,0 +1,18 @@
+#include <math.h>
+#include "trackFinder.hpp"
+
+double getTrack(double latitudeInitial, double longitudeInitial, double latitudeCurrent, double longitudeCurrent)
```
Fberbari: using only 2 points is really noisy because of whatever random gaussian error there is associated with our sensor readings. WHat your'e essentially doing here is calculating a derivative. Numerical methods give us many algorithms to do this, most based off of taylor series, and each with a different error associated. Learned this in a second year class, but it's something you can take a look at now. https://www.dam.brown.edu/people/alcyew/handouts/numdiff.pdf or http://people.bu.edu/andasari/courses/Fall2015/LectureNotes/Lecture7_24Sept2015.pdf are decent intros. What we're really interested in is the backwards difference method (because we only know past gps measurements, not future ones). It's up to you to figure out how many terms to use to give us a reasonable balance between noisiness and a "slow" derivative (one that lags behind what the plane is actually doing by a few iterations). I'd say 4 or 5 terms is probably a pretty good start(consider this loop should run at maybe 50 Hz or so).

Fberbari: I'm not fond of this prototype because you're asking the caller to remember past datapoints for you. Better is to just have them use the most recent data point. You can store the past datapoints in a static array or something inside the function. Also, this way, you can modify the precision of your algorithm (how many terms you use and all) without asking the caller to change his method of calling.

Fberbari (commented):
Good stuff, left a few comments to address, I haven't reviewed all the files just yet though.

Autopilot/PathManager/Inc/trackFinder.hpp
NixonCTChan: Yup, its going to go into the sensorFusion files. I'll add the comments.

Autopilot/PathManager/Src/trackFinder.cpp
NixonCTChan: Oh ok gotcha. Will these two points still be as noisy if we take them after sensorFusion? I was thinking this function should be placed at the end of sensorFusion, where GPS locations are already fused with IMU stuff.

Fberbari: Ya they will. Sensor fusion, even if really good, isn't enough for what you want to do. in 1/50th of a second, the plane moves very little (maybe about 15 cm). Ask Tony, and he'll likely tell you that his best algorithm will give us accuracies that are of that order of magnitude (maybe 5 or 6 cm and that's being realllllllllllllly optimistic). 5 cm off for a 15 cm travel is dozens of degrees of error in the track.
[10:02 AM]
So your algorithm will report the plane is "thrashing", and the control surfaces will act very vilantly to try to correct errors that don't really exist.
The right way to go about this would be to work with Tony to figure out what the error on his algorithm would be (at least if he can give you an order of magnitude, it may be much worse than what I've estimated here) and from there, figure out how many terms you want to use in your backwards difference approximation.

Autopilot/PathManager/Src/AutoSteer.cpp
NixonCTChan: For those who want to know more about the decision that was made on this, basically I contacted a flight instructor about why some people use pitch for speed while others use pitch for altitude. He said its basically interchangeable, it depends on what value you want to control most for each stage of flight. Whichever value you want to control most, use pitch for. For example, on landing, speed is the most important, therefore we use pitch to control speed. 

Anthony and I determined that we are gonna stick with pitch for speed and throttle for altitude throughout the entire flight process for now, but if there are any issues in the future we can change it depending on the scenario. Ex. issues with stalling.

Autopilot/PathManager/Inc/trackFinder.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+
+/*
+    The goal of this function is to return the current track of the plane, given its velocities in the latitude and longitude direction.
+    Given the speeds of the aircraft in the latitude and longitude directions in m/s, this function will output the track in degrees.
+*/
```
Fberbari: It may be nitty gritty, but we should enforce max documentation. See how other prototypes are documented and follow that pattern. You'l need an @param[in] for each of your args to describe what it is, and an @return to describe exactly what your function returns.

Also the 2 sentences in your doc string are exactly the same. You only need 1.

Autopilot/PathManager/Src/trackFinder.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include "trackFinder.hpp"
+
+constexpr double DEG2RAD {3.14159265358979323/180};
+
+double getTrack(double latitudeSpeed, double longitudeSpeed)
+{   
+    double trackInRad = atan2(longitudeSpeed,latitudeSpeed); //the x and y are flipped so that 0 rad is north
+    if(trackInRad < 0)
+    {
+        return (trackInRad/DEG2RAD + 360);
```
Fberbari: division takes much longer than multiplication in general. constexpr are evaluated at compile time, so I'd say you should make it a factor that you multiply rather than one you divide. It's possible that the compiler may optimize this but I'm not sure. May as well make this a habit.

Fberbari (approved):
Nice!

Approved, but please fix the last few little things I mentioned. Apart from that, good to merge!

PR #103, opened 2021-03-31
Update altimeter.hpp

sahil-kale:
As requested by @Pop0097


Fberbari (approved):
And this is fine too.

PR #104, opened 2021-03-31
Updated Gps Header

Pop0097:
For integrating GetSensorData state in path manager.


Fberbari (approved):
Yep, this is fine. May need modification as this driver is developed, but it will do the job atm

PR #105, opened 2021-03-31
Update gps.hpp

Pop0097:


Fberbari (approved):
For anyone wondering, reason this has to be static is that the caller needs to be able to get a singleton without actually already having an instantiated object. So the method that delivers the object needs to be static.

PR #106, opened 2021-03-31
Add new sensor fusion header

antimonyB:


Autopilot/AttitudeManager/Inc/AttitudeDatatypes.hpp
```diff
@@ -44,5 +44,28 @@ struct Airspeed_Data_t
...
+    float groundSpeed; // in m/s
+    int altitude; // in m
+    int heading; // in degrees. Should be between 0-360 at all times, but using integer just in case
+    int numSatellites;    // 1 Byte
+
+    int sensorStatus; // 0 = no fix, 1 = gps fix, 2 = differential gps fix (DGPS) (other codes are possible)
+    bool dataIsNew; // true if data has been refreshed since the previous time GetResult was called, false otherwise.
+
+} Gps_Data_t;
+
```
Pop0097: Wait, attitude manager doesn't use GPS, only airspeed and imu.

 This struct should be declared in the pathManager... Could you perhaps declare the struct in the `SensorFusion.hpp` file? I will need to delete it when integrating it with pathManager tho

Also, the typedef is not needed and it makes our struct declarations inconsistent.

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,60 @@
...
+    long double longitude; //Decimal degrees
+    float longSpeed; //m/s
+} SFPositionOutput_t;
+
+const int NUM_KALMAN_VALUES = 6; 
+
+typedef struct {
+    float prevX[NUM_KALMAN_VALUES];
+    float prevP[NUM_KALMAN_VALUES*NUM_KALMAN_VALUES];
+} SFIterationData_t;
```
Pop0097: Please remove the typedefs. It really isn't needed for struct declarations in c++

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Get position data by fusing data from IMU and GPS.
+ * @param[out]    Output  Reference to an output struct for position data.
+ * @param[in]     altimeterdata
+ * @param[in]     gpsdata
+ * @param[in]     imudata
+ * @param[in,out] iterdata Iterative data from previous call which becomes input for next call.
+ */
+SFError_t SF_GetPosition(SFPositionOutput_t *Output, Altimeter_Data_t *altimeterdata, Gps_Data_t *gpsdata, IMU_Data_t *imudata, SFIterationData_t *iterdata);
```
Pop0097: Okay

So by the looks of it, both path manager and AttitudeManager need access to IMU_Data_t. So it probs makes sense to make an abstraction that contains all of the sensor structs instead of storing them in `attitudeDatatypes.hpp`

Pop0097: Also, how will initializing the `SFIterationData_t` structs work?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
...
+} SFAttitudeOutput_t ;
+
+//The SF_Get functions ensure SensorFusion does not have access to the sensor drivers
+/**
+ * Get attitude data by fusing data from IMU and airspeed sensors.
+ * @param[out]  Output  Reference to an output struct for attitude data.
+ * @param[in]   imudata
+ * @param[in]   airspeeddata
+ */ 
+SFError_t SF_GetAttitude(SFAttitudeOutput_t *Output, IMU_Data_t *imudata, Airspeed_Data_t *airspeeddata);
```
Pop0097: Just confirming, all of this will be deleted?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
...
+    float IMUroll, IMUpitch, IMUyaw; //rad
+    float IMUrollrate, IMUpitchrate, IMUyawrate; //rad/s
+    float Airspeed; //m/s
+    float altitude; //m
+    float altSpeed; //m/s
+    long double latitude; //Decimal degrees
+    float latSpeed; //m/s
+    long double longitude; //Decimal degrees
+    float longSpeed; //m/s
+};
```
Pop0097: Should also have heading in here :))

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
+/**
+ * Sensor Fusion - Converts raw sensor data into human readable formatted structs
+ * Author: Lucy Gong, Dhruv Rawat, Anthony Bertnyk
+ */
+
+#include "AttitudeDatatypes.hpp"
+#include "fetchSensorMeasurementsMode.hpp"
```
Pop0097: What exactly are you including from "fetchSensorMeasurementsMode.hpp"? 

Asking because Chief will be working on a PR that will axe this module (since SF will be collecting data the fetchSensorMeasurements module is redundant)

Pop0097: Also good job not including the sensor header files here!

Pop0097 (commented):
Looking neat! Some requests and questions on my end

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Initiate sensor fusion.
+ */ 
+SFError_t SF_Init(void);
+
+/**
+ * Get fused sensor data.
+ * @param[out]    output  Reference to an output struct for fused data.
+ */ 
+SFError_t SF_GetResult(SFOutput_t *output);
```
Pop0097: Something to consider: 

- It may be better to split SF_execute() functions such that each one executes one specific algorithm. That way Pathmanager doesn't run the Madgwick algorithm and all. 
- Perhaps splitting the SFoutput_t struct into two different structs so AttitudeManager and PathManager only see the sensor data that they care about. You can call them SFAttitudeOutput_t and SFPathOutput_t

Autopilot/Inc/SensorFusion.hpp
antimonyB: Yes

Autopilot/Inc/SensorFusion.hpp
antimonyB: Right, I can remove the include now that SF will collect the data directly instead.

Autopilot/Inc/SensorFusion.hpp
antimonyB: The Kalman filter can actually make use of output from the Madgwick algorithm so I was thinking of combining them to run one after the other. But since Madgwick can run on its own it might be worth splitting them anyways.

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+// -1 = FAILED
+// 0 = SUCCESS
+// 1 = Old Data
+struct SFError_t{
+    int errorCode;
+};
+
+struct SFOutput_t {
+    float IMUroll, IMUpitch, IMUyaw; //rad
```
Fberbari: nit: Can we actually change these names ? The caller doesen't really care that these things come from an IMU. roll, pitch, yaw is fine, same with rates.

Autopilot/Inc/SensorFusion.hpp
Fberbari: heading == yaw as long as it's measured as an azimuth

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
...
+    long double latitude; //Decimal degrees
+    float latSpeed; //m/s
+    long double longitude; //Decimal degrees
+    float longSpeed; //m/s
+};
+
+/**
+ * Initiate sensor fusion.
+ */ 
+SFError_t SF_Init(void);
```
Fberbari: Can init fail ? Why are we returning an error struct?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,62 @@
...
+struct SFError_t{
+    int errorCode;
+};
+
+struct SFOutput_t {
+    float IMUroll, IMUpitch, IMUyaw; //rad
+    float IMUrollrate, IMUpitchrate, IMUyawrate; //rad/s
+    float Airspeed; //m/s
+    float altitude; //m
+    float altSpeed; //m/s
```
Fberbari: altspeed ? What's this ?

Autopilot/Inc/SensorFusion.hpp
Fberbari: Should weigh in here. There's a problem with this design because path manager and attitude manager are both calling execute at different rates. So you end up in this case where execute would no longer get called at regular intervals.

I didn't anticipate that the measurements would be so coupled though, so it's a pain to share data between path and attitude.

Let's talk about this a bit, cuz it's an important consideration, but one thing that can be done is sensor fusion can live in its own thread, and just compute things at whatever constant interval you like (probs as fast as the fastest sensor). And path manager/attitude manager, rather than calling a function that actually computes measurements, can call functions that retrieve the latest measurements.

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+struct SFError_t{
+    int errorCode;
+};
+
+struct SFOutput_t {
+    float roll, pitch, yaw; //rad
+    float rollRate, pitchRate, yawRate; //rad/s
+    float airspeed; //m/s
+    float altitude; //m
+    float altitudeSpeed; //m/s
```
Fberbari: rateOfClimb ?

Fberbari: rateOfClimb ?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+    long double latitude; //Decimal degrees
+    float latitudeSpeed; //m/s
+    long double longitude; //Decimal degrees
+    float longitudeSpeed; //m/s
+};
+
+/**
+ * Initiate sensor fusion.
```
Fberbari: nit: Initialize

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+ */ 
+void SF_Init(void);
+
+/**
+ * Get fused sensor data.
```
Fberbari: So this is just the getter function, right ? It doesn't actually "do" anything ? You should instruct users about how to use each function in your module, including saying things like "it's safe to call this as often as you want and non synchronously".

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+} SFAttitudeOutput_t ;
+
+//The SF_Get functions ensure SensorFusion does not have access to the sensor drivers
+/**
+ * Get attitude data by fusing data from IMU and airspeed sensors.
+ * @param[out]  Output  Reference to an output struct for attitude data.
+ * @param[in]   imudata
+ * @param[in]   airspeeddata
+ */ 
+SFError_t SF_GetAttitude(SFAttitudeOutput_t *Output, IMU_Data_t *imudata, Airspeed_Data_t *airspeeddata);
```
Fberbari, line 67: is this meant to be the function called at a regular interval in its own thread ? The prototype doesn't seem to fit with that. Some info at the top of the function about how to call it, and what it;s meant to do would be real helpful.

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+#define SENSORFUSION_HPP
+
+// -1 = FAILED
+// 0 = SUCCESS
+// 1 = Old Data
+struct SFError_t{
+    int errorCode;
+};
+
+struct SFOutput_t {
```
Pop0097, line 21: Would it be possible to add a heading field. You can refer to the GPS return struct for this

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+//RTOS stuff that the cpp files need
+#ifdef TARGET_BUILD
+
+extern "C"
+{
+#include "cmsis_os.h"
+}
+
+const char SF_MAIL_Q_SIZE = 1;
```
Pop0097: Can use `#define` here

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+//Set up a mail queue for getting data from sensor fusion
+extern osMailQDef(SFMailQ, SF_MAIL_Q_SIZE, SFOutput_t);
+extern osMailQId SFMailQ;
```
Pop0097: Should the declaration for the osMailQId should come before osMailQDef is called?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Initiate sensor fusion.
+ */ 
+void SF_Init(void);
+
+/**
+ * Get fused sensor data.
+ * @param[out]    output  Reference to an output struct for fused data.
+ */ 
+SFError_t SF_GetResult(SFOutput_t *output);
```
Pop0097: I feel like the following should be added: (Feel free to disagree)

- A static instance of the SFOutput_t struct that will be populated with the latest sensor fusion calculations (maybe this should be placed in the .cpp file...)
- A new function called SF_GenerateNewResults(), or some other name, that will be called by the RTOS. This will collect sensor data (maybe by using helper functions) and generate new SFOutput_t data

Now, SF_GetResult() will only return the current values stored in the static SFOutput_t struct declared either in this file or the .cpp file.

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+extern "C"
+{
+#include "cmsis_os.h"
+}
+
+const char SF_MAIL_Q_SIZE = 1;
+
+//Set up a mail queue for getting data from sensor fusion
+extern osMailQDef(SFMailQ, SF_MAIL_Q_SIZE, SFOutput_t);
```
Fberbari: You don't actually need a queue here. Your module will be called by multiple tasks, yes, but in your case, there is a far easier way to share data. Each task in FreeRtos gets its own stack, but the heap is shared by all tasks, and since FreeRtos doesen't deal with virtual memory (it's all physical), then static data is also shared by all tasks.

What this means to you is that if you declare a  static or global variable in your .cpp file, it will be shared by all tasks (because statics/globals live in static memory). A variable declared on the heap would also be able to be shared (but that's not as clean of a solution cuz you gotta pass pointers around).

Autopilot/Inc/SensorFusion.hpp
Fberbari: yes, static instance should live in the .cpp

Autopilot/Inc/SensorFusion.hpp
Pop0097: This function will be deleted. It's there for build purposes right now I think

Autopilot/Inc/SensorFusion.hpp
antimonyB: Would that prevent multiple tasks from accessing the output struct at once? Maybe a queue is overkill, but I am concerned that sensor fusion would start writing to the static/global variable, but partway through another task tries to access it and receives incomplete/broken data.

Fberbari: The way to prevent that is with a Mutex. Mutexes ensure that only one thread can access the ressource at once. 

https://www.freertos.org/Real-time-embedded-RTOS-mutexes.html#:~:text=accompanying%20example%20projects.-,FreeRTOS%20Mutexes,include%20a%20priority%20inheritance%20mechanism.&text=When%20used%20for%20mutual%20exclusion,('take')%20the%20token.

You might like to look deeper into this as well, cuz one optimization is that a thread that's reading doesen't need to wait for another thread that's reading, only for a thread that's writing, so there may be a way to have all the reading threads to all share the token somehow.

This should all live in the cpp file though.

Autopilot/Inc/SensorFusion.hpp
Fberbari: This will be chopped since no need for queue, but for reference, #define is not the best way to do things in c++, since it's not typesafe. constexpr is the name of the game for this kind of thing

Pop0097 (commented):
Looks good to me. Will see what Anthony has to say!

Fberbari (approved):
Approved. Good stuff

PR #107, opened 2021-03-31
Attitude manager now accepts throttle percentage and rudderPercentage…

Fberbari:
… as inputs from path manager. I've changed the intercept file to reflect that.


NixonCTChan (approved):
Great!

PR #108, opened 2021-04-01
Auto Takeoff

NixonCTChan:
- This draft includes the previous landing stuff
- Added takeoff states, functions, and tests
- Added reset variables state and the reset function
- Added passby structure to waypoint manager (some dependency issues come up if I include it in the landingTakeoff file, should I keep it somewhere else?)
- Added passby functionality to attitude manager


Autopilot/Test/Src/LandingTakeoffManager/Test_Takeoff.cpp
```diff
@@ -0,0 +1,60 @@
...
+TEST(TakeoffFunctions, Test_createTakeoffWaypoint){
+    /***SETUP***/
+    _PathData desiredTakeoffPoint;
+    /***STEPTHROUGH***/
+    desiredTakeoffPoint = TakeoffManager::createTakeoffWaypoint(1,2,3, 45);
+    /***ASSERTS***/
+    EXPECT_NEAR(desiredTakeoffPoint.latitude, 1.0006353018, 0.000001);
+    EXPECT_NEAR(desiredTakeoffPoint.longitude, 2.0006352019, 0.000001);
+    EXPECT_NEAR(desiredTakeoffPoint.altitude, 3, 0.000001);
+}
```
Pop0097: Need newline

Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -92,11 +92,30 @@ void PIDloopMode::execute(attitudeManager* attitudeMgr)
...
+    } 
+    if(pathManagerOutput.passbyData.rudderPassby)
+    {
+        _PidOutput.rudderPercent = pathManagerOutput.passbyData.rudderPercent;
+    } 
+    if(pathManagerOutput.passbyData.throttlePassby)
+    {
+        _PidOutput.throttlePercent = pathManagerOutput.passbyData.throttlePercent;
+    } 
+
```
Pop0097: imo it would be better to do if else statements here. No need to execute the PID if we are just going to override the value anyway yknow?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+constexpr static double DISTANCE_OF_LANDING {10}; //in meters
+constexpr static double METERS_PER_DEG_LAT {111320.0};
+constexpr static double ANGLE_OF_LANDING {5}; //in degrees
+constexpr static int STALL_SPEED_NO_PACKAGE {2}; //in m/s
+constexpr static int STALL_SPEED_WITH_PACKAGE {4}; //in m/s
+constexpr static double PI {3.14159265358979323846};
+constexpr static int FLARE_ALTITUDE {2}; //in meters
+constexpr static double DECRAB_ALTITUDE {0.7}; //in meters
+constexpr static double TOUCHDOWN_ALTITUDE {0.05}; //in meters
```
Pop0097, line 27: Curious, why are you not using `#define` here?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
...
+        */
+        static double slowFlightSpeed(bool ifPackage);
+
+        /*
+            This function takes the coordinates of the landing spots and direction of landing
+            Creates stopping points, aiming point, and intersection point
+        */
+        static _LandingPath createSlopeWaypoints(Telemetry_PIGO_t input, double currentAltitude);
+
+};
```
Pop0097: I would strongly advise against using inheritance here. Instead, make a private class member that is an instance of the waypoint manager.

In good OOP design, the derived class should be a specialization of the parent class (the design principle is called Liskov's Principle of Substitutability). I don't think this matches that case... @Fberbari what do you think?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    _PathData stoppingPoint;
+    _PathData aimingPoint;
+    _PathData intersectionPoint;
+};
+
+struct _LandingInitialData
+{
+    double stoppingLatitude;
+    double stoppingLongitude;
```
Pop0097: should be `long double` to keep consistency with other structs

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
...
+            This function takes the horizontal position of the aircraft and determines the desired altitude
+            Projects the position to the vertical plane relative to the aimingPoint, intersectionPoint, and stoppingPoint
+            Takes the one of the horizontal coordinates, plug it into the equation of the slope, get an altitude
+        */
+        static double changingAltitude(_WaypointManager_Data_In input, _PathData aimingPoint, _PathData intersectionPoint, _PathData stoppingPoint);
+
+        /*
+            This function sets the desired altitude so that the throttle turns off
+        */
+        static double throttleOff(void);
```
Pop0097: How does it set the desired altitude if it doesn't take in a parameter?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
...
+/************************
+    TAKEOFF SECTION
+************************/
+
+constexpr static double EXIT_TAKEOFF_ALTITUDE {50}; //in meters, difference in altitude
+constexpr static double ROTATION_SPEED_NO_PACKAGE {4}; //in meters per second
+constexpr static double ROTATION_SPEED_WITH_PACKAGE {6}; //in meters per second
+constexpr static double CLIMB_SPEED_NO_PACKAGE {5}; // in meters per second
+constexpr static double CLIMB_SPEED_WITH_PACKAGE {7}; // in meters per second
+constexpr static double DISTANCE_OF_TAKEOFF {100}; // in meters, just used to approximate the horizontal location
```
Pop0097: Why are we using `constexpr` instead of preprocessors?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
...
+        /*
+            This function returns the desired climb speed to be maintained throughout the climb
+        */
+        static double desiredClimbSpeed(double wind, bool ifPackage);
+        /*
+            This function returns a waypoint that the plane should follow horizontally
+            This waypoint does not have a vertical component, as max throttle is kept throughout the climb
+        */
+        static _PathData createTakeoffWaypoint(double currentLatitude, double currentLongitude, double currentAltitude, double takeoffDirection);
+};
```
Pop0097: Same point as before. But now that I look at it, you may need a constructor to initialize the waypoint manager thinking about it...

Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -0,0 +1,44 @@
...
+    float stoppingDirectionHeading;
+    float stoppingLongitude;
+    float stoppingLatitude;
+    float stoppingAltitude;
+    float takeoffDirectionHeading;
+    // Struct for stopping point gps coordinates
+
+};
+
+#endif
```
Pop0097: new line

Autopilot/AttitudeManager/Inc/GetFromPathManager.hpp
```diff
@@ -2,14 +2,15 @@
...
-
+#include "landingTakeoffManager.hpp"
 #ifndef GET_FROM_PATH_MANAGER_HPP
 #define GET_FROM_PATH_MANAGER_HPP
 
 struct PMCommands{
 	float roll, pitch;	// commanded orientation (radians)
 	float rudderPercent;
 	float throttlePercent;
+	_PassbyControl passbyData;
```
Pop0097, line 15: LandingTakeoffManasger.hpp doesn't have a struct called `_PassbyControl`... where is this struct coming from?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -18,7 +18,27 @@ struct _WaypointManager_Data_In {
...
+    bool rollPassby;
+
+    double rudderPercent;
+    bool rudderPassby;
+
+    double pitchPercent;
+    bool pitchPassby;
+
+    double throttlePercent;
+    bool throttlePassby;
```
Pop0097: Would it be better to put this struct in `pathDatatypes.hpp`. I see no point putting it here if waypoint manager will do nothing with it

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -61,12 +64,59 @@ class sensorFusion : public pathManagerState
         void execute(pathManager* pathMgr);
         void exit(pathManager* pathMgr) {(void) pathMgr;}
         static pathManagerState& getInstance();
+        static _WaypointManager_Data_In sensorInput; //must be filled out
```
Pop0097: Ideally parameters should not be made public

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -61,12 +64,59 @@ class sensorFusion : public pathManagerState
...
+    public:
+        void enter(pathManager* pathMgr) {(void) pathMgr;}
+        void execute(pathManager* pathMgr);
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+    private:
+        resetVariables() {}
+        resetVariables(const resetVariables& other);
+        resetVariables& operator =(const resetVariables& other);
+        void resetPassby(_PassbyControl* controlDetails){controlDetails->rollPassby = controlDetails->pitchPassby = controlDetails->throttlePassby = controlDetails->rudderPassby = false;}
```
Pop0097, line 86: I've never seen this type of syntax before... what's it called?

Pop0097: Also if there is a more readable way to do this, that would be best (also it may be worth defining the function in pathStateClasses.cpp

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -61,12 +64,59 @@ class sensorFusion : public pathManagerState
...
+    public:
+        void enter(pathManager* pathMgr) {(void) pathMgr;}
+        void execute(pathManager* pathMgr);
+        void exit(pathManager* pathMgr) {(void) pathMgr;}
+        static pathManagerState& getInstance();
+        static WaypointManager takeoffPath;
+        static _PathData * pathArray[1];
+        static _PathData * currentLocation;
+        static _WaypointStatus waypointStatus;
+        static _PathData takeoffPoint;
```
Pop0097: Data parameters usually should always be made private.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,102 @@
+#pragma once
+
```
Pop0097, line 2: Curious, why are we putting both the landing and takeoff manager In the same file? Would it make sense to break them into separate files?

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,184 @@
+#include "waypointManager.hpp"
+#include "landingTakeoffManager.hpp"
+#include <math.h>
+#include "vectorClass.hpp"
```
Pop0097: Don't need to include waypointManager and vectorClass as they are already included in landingTakeoffManager

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+    Vector3D iPoint(intersectionPoint.longitude, intersectionPoint.latitude, intersectionPoint.altitude); //intersection point
+    Vector3D sPoint(stoppingPoint.longitude, stoppingPoint.latitude, stoppingPoint.altitude); //stopping point
+    Vector3D cPoint(input.longitude, input.latitude, input.altitude); //current point
+    Vector3D projectedPoint;
+
+    //calculating the vectors
+    Vector3D vectorAI(iPoint.x - aPoint.x, iPoint.y - aPoint.y, iPoint.z - aPoint.z);
+    Vector3D vectorAS(sPoint.x - aPoint.x, sPoint.y - aPoint.y, sPoint.z - aPoint.z);
+    Vector3D vectorAC(cPoint.x - aPoint.x, cPoint.y - aPoint.y, cPoint.z - aPoint.z);
+    Vector3D normal;
```
Pop0097: the variable `normal` is blue here... that kinda worries me because git is interpreting it as something. Maybe change the name to normalVector?

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+{
+    //slow flight speed calculation for both package and no package scenarios
+    if(ifPackage)
+    {
+        return STALL_SPEED_WITH_PACKAGE + 2;
+    }
+    else 
+    {
+        return STALL_SPEED_NO_PACKAGE + 2;
+    }
```
Pop0097, line 62: would it make sense to make it slightly faster (to account for some more error)?

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,184 @@
+#include "waypointManager.hpp"
```
Pop0097: Yeah, looking at your implementations, these classes should not inherit WaypointManager. If you could instead create an instance of waypoint manager in the classes that would be a lot better deign-wise

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -55,14 +55,11 @@ pathManagerState& getSensorData::getInstance()
...
-    if(isError)
+    if(pathMgr->isError)
     {
         pathMgr -> setState(fatalFailureMode::getInstance());
     }
-    else
-    {
-        pathMgr -> setState(cruisingState::getInstance());
-    }
+    pathMgr->setState(sensorFusion::getInstance());
```
Pop0097: Keep the else{} statement here for consistency please

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
+    {
+        //requires data structure that dhruv wants to use 
+        landingPath = WaypointManager(0.0,0.0);
+        path = LandingManager::createSlopeWaypoints(getFromTelemetry::telemetryInput, sensorFusion::sensorInput.altitude);
+
+        //creating waypoints 
+        pathArray[0] = landingPath.initialize_waypoint(path.intersectionPoint.longitude, path.intersectionPoint.latitude, path.intersectionPoint.altitude, PATH_FOLLOW);
+        pathArray[1] = landingPath.initialize_waypoint(path.aimingPoint.longitude, path.aimingPoint.latitude, path.aimingPoint.altitude, PATH_FOLLOW);
+        pathArray[2] = landingPath.initialize_waypoint(path.stoppingPoint.longitude, path.stoppingPoint.latitude, path.stoppingPoint.altitude, PATH_FOLLOW);
+        currentLocation = landingPath.initialize_waypoint(sensorFusion::sensorInput.longitude, sensorFusion::sensorInput.latitude, sensorFusion::sensorInput.altitude, HOLD_WAYPOINT); //fill in with sensor fusion data
```
Pop0097: If you are defining a hold waypoint, you should call this method:

`_PathData* WaypointManager::initialize_waypoint(long double longitude, long double latitude, int altitude, _WaypointOutputType waypointType, float turnRadius)`

instead

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
 
+/****************************************************************************************************
+
+LANDING STATE FUNCTIONS
+
+****************************************************************************************************/
+
+void landingTransitionStage::execute(pathManager* pathMgr)
+{
+    if(!pathMgr -> madeLandingPoints)
```
Pop0097: Styling, but don't put spaces around `->`

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -1,6 +1,5 @@
 #include "pathStateClasses.hpp"
-
-bool isError; 
+#include "landingTakeoffManager.hpp"
```
Pop0097: Don't need this

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
+        //set made madelandingPoints to true
+        pathMgr->madeLandingPoints = true;
+    }
+
+    //follow the landing waypoints
+    waypointStatus = landingPath.get_next_directions(sensorFusion::sensorInput, &cruisingState::_outputdata);
+
+    //calculating the difference in heading to detect if finished turning (2 differences in heading possible)
+    differenceInHeading1 = getFromTelemetry::telemetryInput.stoppingDirectionHeading - sensorFusion::sensorInput.track;
+    differenceInHeading2 = sensorFusion::sensorInput.track - getFromTelemetry::telemetryInput.stoppingDirectionHeading;
```
Pop0097: I don't like how we are just accessing variables from other classes like this. I would rather you made getter functions since it will prevent people (in the future) from actually changing the values.

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
+
+    //making sure both headings are positive
+    if(differenceInHeading1 < 0){differenceInHeading1 += 360;}
+    if(differenceInHeading2 < 0){differenceInHeading2 += 360;}
+
+    //if the smaller heading is less than 5 degrees, set stage to slope
+    if(differenceInHeading1<differenceInHeading2)
```
Pop0097: space around `<` and other comparative operators

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
+    {
+        if(fabs(differenceInHeading1) <= 5)
```
Pop0097: Suggestion: you can use the && operator to get rid of the nested if statements

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
+
+****************************************************************************************************/
+
+void landingTransitionStage::execute(pathManager* pathMgr)
+{
+    if(!pathMgr -> madeLandingPoints)
+    {
+        //requires data structure that dhruv wants to use 
+        landingPath = WaypointManager(0.0,0.0);
+        path = LandingManager::createSlopeWaypoints(getFromTelemetry::telemetryInput, sensorFusion::sensorInput.altitude);
```
Pop0097: if you are creating a new waypoint manager you can actually make a constructor for LandingManager. So getting rid of the inheritance should not be too complex

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +164,287 @@ pathManagerState& fatalFailureMode::getInstance()
...
+    if(sensorFusion::sensorInput.altitude <= (TOUCHDOWN_ALTITUDE + getFromTelemetry::telemetryInput.stoppingAltitude)) //altitude is 5 cm or less/ultrasonic sensor sensed 5cm or less
+    {
+        pathMgr->stage = TOUCHDOWN;
+    }
+    else
+    {
+        //align heading with landing direction
+        cruisingState::_outputdata.desiredHeading = getFromTelemetry::telemetryInput.stoppingDirectionHeading;
+        //retrieving desired slow flight speed
+        cruisingState::_outputdata.desiredSpeed = LandingManager::slowFlightSpeed(getFromTelemetry::telemetryInput.packageOnBoard);
```
Pop0097: Why are you accessing crusingState parameters?

Pop0097 (commented):
Looking good so far!

Most of my comments have to do with design and styling.

Autopilot/Test/Src/LandingTakeoffManager/Test_Takeoff.cpp
NixonCTChan: gotcha

Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
NixonCTChan: true, fixed!

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: I was talking to Anthony about this, he recommended constexpr to define constants at compile time. Is there any advantage to use define instead?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: Yeah now that I think about it again, thats true. I'll remove the inheritance completely. Since its just a class of functions, there doesnt need to be any connections between waypoint manager and landing takeoff stuff.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: gotcha

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: Yeah forgot to remove these functions, their functionality is being replaced with the passbyControl structure.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: Answered above

Autopilot/PathManager/Inc/waypointManager.hpp
NixonCTChan: yup will do

Autopilot/PathManager/Inc/pathStateClasses.hpp
NixonCTChan: some of these variables may need to be accessed and modified by other states. Having getters and setters for each variable would get very cluttered, which is why I'm making them public. Do you want to do it with getters and setters?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
Pop0097: https://stackoverflow.com/questions/42388077/constexpr-vs-macros/42388188

Covers the differences well. I agree we should stick to constexpr :))

Autopilot/PathManager/Inc/pathStateClasses.hpp
NixonCTChan: http://courses.cms.caltech.edu/cs11/material/cpp/donnie/cpp-ops.html
refer to assignment operator chaining. I would be able to do it another way, just wanted to see if you guys were ok with chaining operators like this.

Autopilot/PathManager/Inc/pathStateClasses.hpp
Pop0097: I personally would rather we use getters and setters. @Fberbari what do you think?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: some of their functions are really similar, and it provides us one file where we can tune the important speeds. Ex. approach speed, climb speed. I don't see why we need to separate them into two files

Autopilot/PathManager/Src/landingTakeoffManager.cpp
NixonCTChan: had that issue before, it doesn't actually do anything but sure, I'll change it to normalVector.

Autopilot/PathManager/Inc/pathStateClasses.hpp
Pop0097: Damn, I didn't know we could do this. Thanks for sharing! Maybe put a little comment linking the website and saying you're doing operating chaining?

Autopilot/PathManager/Src/landingTakeoffManager.cpp
NixonCTChan: these values will need to be tuned when we're simulating the landing takeoff stuff, they're not set in stone.

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: Answered above

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: There doesn't need to be a constructor, landing and takeoff classes are just a class of functions.

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: Since coordinated turns would need to be pulling instructions from a wide variety of states, I was thinking to modify one variable for all the control states. Now that I think about it, I think it's better for all states to modify an input variable in coordinate turns instead of cruising. What do you think?

Autopilot/PathManager/Inc/pathStateClasses.hpp
NixonCTChan: Yup, I'll put it down.

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -11,6 +11,7 @@
 #include <stdlib.h>
 #include <math.h>
 #include <cstdint>
+#include "pathDatatypes.hpp"
```
Pop0097, line 19: Will waypoint manager be using any of the structs in here? If not, then no need to include it

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -202,13 +204,11 @@ void landingTransitionStage::execute(pathManager* pathMgr)
...
+    if(differenceInHeading1 < differenceInHeading2 && fabs(differenceInHeading1) <= 5)
     {
-        if(fabs(differenceInHeading1) <= 5)
-        {
-            //set enum to slope state
-            pathMgr->stage = SLOPE;
-        }
+        pathMgr->stage = SLOPE;
     }
     else
```
Pop0097: use `else if ()` here instead.

Pop0097 (commented):
Nice improvements. 

Some comments still have to be addressed (like public class members and all, but overall fix seems good so far).

Autopilot/PathManager/Inc/waypointManager.hpp
NixonCTChan: yeah, the passby control is in the waypoint manager out struct. Should I make it separate?

NixonCTChan:
Changes made:
- added landing and takeoff state transition tests
- added enum in CT to determine what instructions to pull from
- added functionality to telemetry input for beginLanding and beginTakeoff
- created landing and takeoff I/O structs, and the translation functions for those
- created variable definitions at the top of pathStateClasses.cpp to fix undefined reference error
- integrated cruising state I/O functions into the enums for both resetVariables and CT

Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -0,0 +1,75 @@
...
+    Telemetry_Waypoint_Data_t waypoints[100]; // Somehow need to get PATH_BUFFER_SIZE here...
+    Telemetry_Waypoint_Data_t homebase;
+
+    // landing and takeoff 
+    bool beginLanding;
+    bool beginTakeoff;
+    float stoppingDirectionHeading;
+    float stoppingLongitude;
+    float stoppingLatitude;
+    float stoppingAltitude;
```
Pop0097, line 50: I'm assuming all of this will be inside of the stopping point gps struct?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,132 @@
...
+    double rudderPercent;
+    bool rudderManual;
+
+    double elevatorPercent;
+    bool elevatorManual;
+
+    double throttlePercent;
+    bool throttleManual;
+};
```
Pop0097: This struct is a carbon copy of _PassbyControl and I couldn't seem to find a place where it is being used. Would it be worth to delete it?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,132 @@
+#pragma once
+
+
+#include "waypointManager.hpp"
+#include "vectorClass.hpp"
+#include "AutoSteer.hpp"
+#include "SensorFusion.hpp"
```
Pop0097, line 10: I would #inlude pathDatatypes.hpp since you are referencing _PassbyControl here :))

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+
+class LandingManager{
+    
+    public:
+        /*
+            This function takes the horizontal position of the aircraft and determines the desired altitude
+            Projects the position to the vertical plane relative to the aimingPoint, intersectionPoint, and stoppingPoint
+            Takes the one of the horizontal coordinates, plug it into the equation of the slope, get an altitude
+        */
```
Pop0097: Love the comments. Little room for improvement: For comments on methods, this structure is usually a bit better:

```
        /**
          *  This function takes the horizontal position of the aircraft and determines the desired altitude
          * Projects the position to the vertical plane relative to the aimingPoint, intersectionPoint, and stoppingPoint
          *  Takes the one of the horizontal coordinates, plug it into the equation of the slope, get an altitude
          * 
          * @param WRITE ONE OF THESE STATEMENTS FOR EACH PARAMETER
          * 
          * @return WHAT IS BEING RETURNED? 
        */
```

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -90,16 +231,79 @@ pathManagerState& cruisingState::getInstance()
...
+            //should default to cruising
+        
+            turnInput.currentHeadingTrack = sensFusionOutput->track; // Gets track;
+            turnInput.desiredHeadingTrack = waypointOutput->desiredTrack;
+            turnInput.accY = imudata->accy;
+            
+            // Set up the compute altitude and airspeed function input
+            altAirspeedInput.currentAltitude = sensFusionOutput->altitude;
+            altAirspeedInput.desiredAltitude = waypointOutput->desiredAltitude;
+            altAirspeedInput.currentAirspeed = sensFusionOutput->airspeed;
```
Pop0097: Would it make more sense to use if statements here? Just wondering since half of the conditions have the same body

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -1,11 +1,87 @@
...
+IMU_Data_t sensorFusion::_imudata;
+CoordinatedTurnAttitudeManagerCommands_t coordinateTurnElevation::_rollandrudder;
+AltitudeAirspeedCommands_t coordinateTurnElevation::_pitchandairspeed; 
+AttitudeData commsWithAttitude::_receivedData;
+
+// Landing and Takeoff Variables (some stages DO NOT need waypoint inputs, it's not a mistake that some are missing)
+_LandingTakeoffInput takeoffRollStage::input;
+_LandingTakeoffOutput takeoffRollStage::output;
+_WaypointManager_Data_Out takeoffRollStage::waypointOutput;
+_PathData * takeoffRollStage::currentLocation;
```
Pop0097, line 23: You are using the identifiers "waypointOutput", etc. a lot. Maybe add the names of the states they belong to? That way it'll be a bit more obvious to a reader

This should especially be done to public members

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+
+    //if the smaller heading is less than 5 degrees, set stage to slope
+    if(differenceInHeading1 < differenceInHeading2 && fabs(differenceInHeading1) <= 5)
+    {
+        pathMgr->stage = SLOPE;
+    }
+    else if(fabs(differenceInHeading2) <= 5)
+    {
+        pathMgr->stage = SLOPE;
+    }
```
Pop0097: You can use the || operand to make this one if statement instead of two

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+
+    //follow the landing waypoints
+    waypointStatus = landingPath.get_next_directions(waypointInput, &waypointOutput);
+
+    //translate waypoint commands into landing output structure
+    output = LandingManager::translateWaypointCommands(waypointOutput);
+
+    //calculating the difference in heading to detect if finished turning (2 differences possible)
+    differenceInHeading1 = input.telemetryData.stoppingDirectionHeading - input.sensorOutput.track;
+    differenceInHeading2 = input.sensorOutput.track - input.telemetryData.stoppingDirectionHeading;
```
Pop0097: Do we need to make these static members of the state? by the looks of it we aren't referring to past values that they held and other states aren't using them.

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+        //requires data structure that dhruv wants to use 
+        path = LandingManager::createSlopeWaypoints(input.telemetryData, input.sensorOutput.altitude);
+
+        //creating waypoints 
+        pathArray[0] = landingPath.initialize_waypoint(path.intersectionPoint.longitude, path.intersectionPoint.latitude, path.intersectionPoint.altitude, PATH_FOLLOW);
+        pathArray[1] = landingPath.initialize_waypoint(path.aimingPoint.longitude, path.aimingPoint.latitude, path.aimingPoint.altitude, PATH_FOLLOW);
+        pathArray[2] = landingPath.initialize_waypoint(path.stoppingPoint.longitude, path.stoppingPoint.latitude, path.stoppingPoint.altitude, PATH_FOLLOW);
+        currentLocation = landingPath.initialize_waypoint(input.sensorOutput.longitude, input.sensorOutput.latitude, input.sensorOutput.altitude, HOLD_WAYPOINT, 20); //fill in with sensor fusion data
+        //initializing flight path
+        waypointStatus = landingPath.initialize_flight_path(pathArray, 3, currentLocation);
```
Pop0097: Amazing job using the API! I hope it was easy to use!

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+        waypointInput.track = input.sensorOutput.track;
+
+        //aligning horizontal position using waypointManager get_next_directions
+        landingTransitionStage::waypointStatus = landingTransitionStage::landingPath.get_next_directions(waypointInput, &waypointOutput);
+        //translate waypoint ouput data into landing data
+        output = LandingManager::translateWaypointCommands(waypointOutput);
+        //retrieving desired altitude for slope state and setting it 
+        output.desiredAltitude = LandingManager::changingAltitude(input.sensorOutput,landingTransitionStage::path.aimingPoint, landingTransitionStage::path.intersectionPoint, landingTransitionStage::path.stoppingPoint);
+        //retrieving desired speed for approach speed and setting it
+        output.desiredAirspeed = LandingManager::approachSpeed(pathMgr->isPackage);
```
Pop0097: Adding some spacing like you did in landingFlareStage::execute() will make this more readable!

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+        output.controlDetails.throttlePassby = true;
+        output.controlDetails.throttlePercent = 0;
+
+        //maintaing speed for flare attitude
+        output.desiredAirspeed = LandingManager::slowFlightSpeed(pathMgr->isPackage);
+    }   
+
+    if(landingTransitionStage::waypointStatus == INVALID_PARAMETERS)
+    {
+        pathMgr -> isError = true;
```
Pop0097: Please remove spaces around `->`. There are a few places where you did this

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+
+        //throttle off
+        output.controlDetails.throttlePassby = true;
+        output.controlDetails.throttlePercent = 0;
+
+        //maintaing speed for flare attitude
+        output.desiredAirspeed = LandingManager::slowFlightSpeed(pathMgr->isPackage);
+    }   
+
+    if(landingTransitionStage::waypointStatus == INVALID_PARAMETERS)
```
Pop0097: Probs would use waypointStatus != WAYPOINT_SUCCESS instead!

Pop0097: (There are a few error codes, but only one success code)

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+    if(input.sensorOutput.altitude <= (TOUCHDOWN_ALTITUDE + input.telemetryData.stoppingAltitude)) //altitude is 5 cm or less/ultrasonic sensor sensed 5cm or less
+    {
+        pathMgr->stage = TOUCHDOWN;
+    }
+    else
+    {
+        //align heading with landing direction
+        output.desiredHeading = input.telemetryData.stoppingDirectionHeading;
+        output.useHeading = true;
+        //retrieving desired slow flight speed
```
Pop0097: I'm a bit confused about where we are using this?

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+    {
+        takeoffPoint = TakeoffManager::createTakeoffWaypoint(input.sensorOutput.latitude,input.sensorOutput.longitude, input.sensorOutput.altitude, input.telemetryData.takeoffDirectionHeading);
+        pathArray[0] = takeoffPath.initialize_waypoint(takeoffPoint.longitude, takeoffPoint.latitude, takeoffPoint.altitude, PATH_FOLLOW);
+        currentLocation = takeoffPath.initialize_waypoint(input.sensorOutput.longitude, input.sensorOutput.latitude, input.sensorOutput.altitude, HOLD_WAYPOINT);
+        waypointStatus = takeoffPath.initialize_flight_path(pathArray, 1, currentLocation);
+
+        pathMgr->madeTakeoffPoints = true;
+    }
+
+    if(input.sensorOutput.airspeed>(TakeoffManager::desiredRotationSpeed(pathMgr->isPackage)))
```
Pop0097: space around > pls

Autopilot/Test/Src/PathManager/Test_PathManager_Fsm.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+	pathMan.setState(resetVariables::getInstance());
+	pathMan.execute();
+
+	/**********************ASSERTS**********************/
+
+	EXPECT_EQ(*(pathMan.getCurrentState()), takeoffClimbStage::getInstance());
+}
+
+TEST (PathManagerFSM, TransitionToCT) {
```
Pop0097: TransitionToCoordinatedTurnElevation

This will make the test names more descriptive!

Autopilot/Test/Src/PathManager/Test_PathManager_Fsm.cpp
```diff
@@ -0,0 +1,433 @@
+/*
+* Author: Dhruv Rawat
+*/
```
Pop0097: Feel free to add your name here!

Autopilot/Test/Src/PathManager/Test_PathManager_Fsm.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+#include "fff.h"
+#include <gtest/gtest.h>
```
Pop0097, line 6: Would it make sense to add tests to see the changes in the enum? Like transition --> slope, etc.?

Autopilot/Test/Src/PathManager/Test_PathManager_Fsm.cpp
```diff
@@ -0,0 +1,433 @@
...
+	/********************DEPENDENCIES*******************/
+	/********************STEPTHROUGH********************/
+
+	pathMan.setState(resetVariables::getInstance());
+	pathMan.execute();
+
+	/**********************ASSERTS**********************/
+
+	EXPECT_EQ(*(pathMan.getCurrentState()), landingTransitionStage::getInstance());
+}
```
Pop0097, line 201: Great job with the unit tests! I especially love how readable they are!

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,204 @@
...
+    Vector3D aPoint(aimingPoint.longitude, aimingPoint.latitude, aimingPoint.altitude); //aiming point
+    Vector3D iPoint(intersectionPoint.longitude, intersectionPoint.latitude, intersectionPoint.altitude); //intersection point
+    Vector3D sPoint(stoppingPoint.longitude, stoppingPoint.latitude, stoppingPoint.altitude); //stopping point
+    Vector3D cPoint(input.longitude, input.latitude, input.altitude); //current point
+    Vector3D projectedPoint;
+
+    //calculating the vectors
+    Vector3D vectorAI(iPoint.x - aPoint.x, iPoint.y - aPoint.y, iPoint.z - aPoint.z);
+    Vector3D vectorAS(sPoint.x - aPoint.x, sPoint.y - aPoint.y, sPoint.z - aPoint.z);
+    Vector3D vectorAC(cPoint.x - aPoint.x, cPoint.y - aPoint.y, cPoint.z - aPoint.z);
```
Pop0097: Maybe add comments saying "Aiming --> Intersection"? 

Comment format
// (starting point) --> (ending Point) 

if you think it isn't needed, feel free to ignore!

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,204 @@
...
+    //finding the x and y components of the rolling distance vector
+    double stoppingDistX = sin(radianDirection) * DISTANCE_OF_LANDING; //5root2
+    double stoppingDistY = cos(radianDirection) * DISTANCE_OF_LANDING; //5root2
+
+    //converting into x and y components in lat and lon
+    double metersPerDegLon = 40075000.0 * cos(input.stoppingLatitude*PI/180)/360.0; //111251.6316
+    double stoppingDistLon = stoppingDistX / metersPerDegLon; //6.355922792e-5
+    double stoppingDistLat = stoppingDistY / METERS_PER_DEG_LAT; //6.352019235e-5
+
+
```
Pop0097: remove two blank lines here pls

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+    else 
+    {
+        return STALL_SPEED_NO_PACKAGE + 2;
+    }
+}
+
+_LandingPath LandingManager::createSlopeWaypoints(Telemetry_PIGO_t input, double currentAltitude)
+{
+    //created comments for testing
```
Pop0097: May want to remove these comments before merging

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,204 @@
...
+    
+    //determining x and y of aiming point
+    double radianDirection = input.stoppingDirectionHeading * PI / 180.0; //PI/4
+
+    //finding the x and y components of the rolling distance vector
+    double stoppingDistX = sin(radianDirection) * DISTANCE_OF_LANDING; //5root2
+    double stoppingDistY = cos(radianDirection) * DISTANCE_OF_LANDING; //5root2
+
+    //converting into x and y components in lat and lon
+    double metersPerDegLon = 40075000.0 * cos(input.stoppingLatitude*PI/180)/360.0; //111251.6316
```
Pop0097: would it make sense to make 40075000.0 a constexpr?

Pop0097 (commented):
Damn! Really good work!

Autopilot/PathManager/Inc/pathDatatypes.hpp
NixonCTChan: do you mean if I'm gonna create a new struct to hold all these in?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: oops sorry, this leaked through when solving the merge conflicts

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: these statements are actually different since they pull commands from each of the different states. If I were to use if statements, there would be just as many cases.

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: most of these are just placeholders for waypoint data that gets outputted, which are then immediately translated to landing and takeoff commands. They are all private members. I'll put a comment that explains it better

Autopilot/Test/Src/PathManager/Test_PathManager_Fsm.cpp
NixonCTChan: it would require some fiddling with the sensor data, but yeah I'll make some.

NixonCTChan: hmm I've tried doing that, but the values that determine if the plane is going to switch enums get updated in the execute function. It's gonna require a lot of sensor measurements and other variables in order to make the state machine think that it should be switching enums.

Autopilot/PathManager/Src/landingTakeoffManager.cpp
NixonCTChan: sure i'll add them, improves readability too.

Autopilot/PathManager/Src/landingTakeoffManager.cpp
NixonCTChan: sure I can do that!

Autopilot/PathManager/Src/pathStateClasses.cpp
Pop0097: Fair enough. Thanks for clearing it up!

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,377 @@ pathManagerState& fatalFailureMode::getInstance()
...
+        output.desiredAltitude = LandingManager::changingAltitude(input.sensorOutput,landingTransitionStage::path.aimingPoint, landingTransitionStage::path.intersectionPoint, landingTransitionStage::path.stoppingPoint);
+        
+        //retrieving desired speed for approach speed and setting it
+        output.desiredAirspeed = LandingManager::approachSpeed(pathMgr->isPackage);
+    }
+    
+    if(!landingTransitionStage::waypointStatus == WAYPOINT_SUCCESS)
+    {
+        pathMgr->isError = true;
+    }
```
Pop0097: Should be `landingTransitionStage::waypointStatus != WAYPOINT_SUCCESS` since it's an integer and not a boolean.

Pop0097: I think you may have done this in other places too

Pop0097 (commented):
Just one little thing from me. @Fberbari what say you?

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -18,72 +28,232 @@ class commsWithAttitude : public pathManagerState
...
+    private:
+        int waypointIDArray[PATH_BUFFER_SIZE];
+        static _LandingTakeoffInput input;
+        static _LandingTakeoffOutput output;
+        static _WaypointManager_Data_In waypointInput;
+        static _WaypointManager_Data_Out waypointOutput;
+        static _PathData * pathArray[3]; //used to translate loaded in path to something the waypoint manager can take as a parameter
+        static _PathData * currentLocation;
+        double differenceInHeading1;
+        double differenceInHeading2;
```
Pop0097: Another thing actually, I don't think we even need these to be member variables for this class. 

Instead, you can declare them in the ::execute() method for this class. That way we aren't populating this class structure with stuff that aren't really being cached or accessed elsewhere.

NixonCTChan: true, I'll make those

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: oops, messed up the precedences. Will fix

Autopilot/.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
Fberbari: Should get rid of this file

Autopilot/AttitudeManager/Inc/GetFromPathManager.hpp
```diff
@@ -2,14 +2,15 @@
  * Gets the commanded orientation and airspeed from the path manager.
  * Author: Anthony Berbari
  */
-
+#include "pathDatatypes.hpp"
```
Fberbari: This should be inside the #ifdef  guard. Reason being that if this file is included recursively, the guard prevents its contents from showing up twice. In practice, I've never seen a case where things are not encapsulated inside the guard

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,163 @@
...
+constexpr static double PI {3.14159265358979323846};
+constexpr static int FLARE_ALTITUDE {2}; //in meters
+constexpr static double DECRAB_ALTITUDE {0.7}; //in meters
+constexpr static double TOUCHDOWN_ALTITUDE {0.05}; //in meters
+constexpr static double LATITUDE_LONGITUDE_CONVERSION_CONSTANT {40075000.0};
+
+struct _LandingPath
+{
+    _PathData stoppingPoint;
+    _PathData aimingPoint;
```
Fberbari: Comments on these things pls (and even a link to your confluence entry). Otherwise it's really intimidating to understand.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,163 @@
...
+    _PathData intersectionPoint;
+};
+
+struct _LandingTakeoffInput
+{
+    Telemetry_PIGO_t telemetryData;
```
Fberbari: Note to myself. These things should be passed by reference and should be read only.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,163 @@
...
+    SFOutput_t sensorOutput;
+};
+
+struct _LandingTakeoffOutput
+{
+    double desiredHeading;
+    double desiredTrack;
+    double useHeading; //pay attention
```
Fberbari: THis comment doesen't mean anything atm.

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -90,16 +231,79 @@ pathManagerState& cruisingState::getInstance()
...
+            LandingManager::translateLTSFCommandsToCoordTurns(landingTransitionStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
+            break;
+        case SLOPE:
+            LandingManager::translateLTSFCommandsToCoordTurns(landingSlopeStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
+            break;
+        case FLARE:
+            LandingManager::translateLTSFCommandsToCoordTurns(landingFlareStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
+            break;
+        case DECRAB:
+            LandingManager::translateLTSFCommandsToCoordTurns(landingDecrabStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
```
Fberbari: I think the only thing changing here is where your'e getting the control input from. So that should be the only thing in the switch statement, the common stuff should just be a single line below the switch statement.

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -90,16 +231,79 @@ pathManagerState& cruisingState::getInstance()
...
+    
     //get elevation and turning data
-    if(isError)
+    //loading in commands data from each of the states
+    switch(pathMgr->stage){
+        case ROLL:
+            LandingManager::translateLTSFCommandsToCoordTurns(takeoffRollStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
+            break;
+        case CLIMB:
+            LandingManager::translateLTSFCommandsToCoordTurns(takeoffClimbStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
```
Fberbari: Climb is being taken care of by the landing manager ? That's very unintuitive

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -120,4 +324,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
 }
 
+/****************************************************************************************************
+LANDING STATE FUNCTIONS
+****************************************************************************************************/
+
+void landingTransitionStage::execute(pathManager* pathMgr)
+{
+    //load in sensor fusion data and telemtry data into input structure
+    input.telemetryData = *(commsWithTelemetry::GetTelemetryIncomingData());
```
Fberbari: Why are we passing these structs by value ? They're read only. They should be passed as constant references or something C++y of the sort.

Fberbari (commented):
Ok, ya, this is a lot. Which is my own fault for neglecting to review this fr some time. But there's some really great stuff here, really great, but it's really overwhelming too. 

There's a couple things I'm nervous about, like the complexity of path manager, which seems to be growing at an insane rate.


Regardless, with 33 files to review, I'm getting really confused about what changes are yours and what's in here because of the rebasing issues. 

Prioritize fixing the rebasing crap, and push again and I'll complete a deep review.

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: the cruising state outputs are accessed completely differently though, that's why I just have the accessing functions in each case statement. I'm not sure how else I can implement this more simply.

Autopilot/PathManager/Src/pathStateClasses.cpp
NixonCTChan: The function for translating both landing and takeoff outputs to coord turn commands lives in the landing manager class. Should I make it a stand alone function?

Fberbari: I'll take a closer look and lyk, this may be the best solution, but it feels weird

Autopilot/PathManager/Src/pathStateClasses.cpp
Fberbari: What I mean is, you can make it so that rather than having all this in each statement

 LandingManager::translateLTSFCommandsToCoordTurns(landingDecrabStage::getControlOutput(), sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
 
 you can have:
 
 ControlInput_t controlInput =  LandingDecrabStage::getControlOutput();
 
 (with the variabe being asigned whatever it needs to be, then at the bottom, you can have a single call to the function with:
 
  LandingManager::translateLTSFCommandsToCoordTurns(controlInput , sensFusionOutput, imudata, &turnInput, &altAirspeedInput);
  
  
 The 2 reasons I note this:
  
  I'm sure you've heard it said that repetition in software is bad. This is a case of that, if for whatever reason, you have to change the names of the other args or of the function or whatever, you'l have to make that change in each case with your current implementation, which is prone to mistakes and is a bit of a pain.
  
  Second reason is that this current implementation forces readers to read every line, playing "find the differences" to figure out what your'e trying to do. That's much more tough than being able to just scan it over.
  
  
  
  
  Is it the biggest issue in the world ? no. But if you get in the habbit of doing little things like this, it'l slowly make you a better developer.

NixonCTChan: ok gotcha, I'll make those changes

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,167 @@
...
+constexpr static int FLARE_ALTITUDE {2}; //in meters
+constexpr static double DECRAB_ALTITUDE {0.7}; //in meters
+constexpr static double TOUCHDOWN_ALTITUDE {0.05}; //in meters
+constexpr static double LATITUDE_LONGITUDE_CONVERSION_CONSTANT {40075000.0};
+
+struct _LandingPath //this structure is used to hold the waypoint dat of the landing path, when it is created in the transition stage
+{
+    _PathData stoppingPoint;
+    _PathData aimingPoint;
+    _PathData intersectionPoint;
```
Fberbari: hate to be nitpicky, but what I really meant was explaining things like what the intersection point is in a few words would make things much more understandable.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+    public:
+        /*
+            This function returns the desired rotation speed limit, where the plane will transition from roll to climb
+            
+            @param bool * ifPackage - this boolean holds if there is a package on board the aircraft
+
+            @return double - this function will return the desired rotation speed for the aircraft
+        */
+        static double desiredRotationSpeed(const bool * ifPackage);
```
Fberbari: As a general rule. Individual built in variables, like bools, ints, and doubles, should be passed by value. Structs should be passed by reference.

Autopilot/PathManager/Inc/pathDatatypes.hpp
Fberbari: I'd be ok with either creating a new struct or not. No big deal either way.

Autopilot/PathManager/Inc/trackFinder.hpp
```diff
@@ -1,7 +1,9 @@
 #pragma once
```
Fberbari: THis file does not need to exist. Track is just pulled from the sensor fusion output. The guts of this file should live inside SensorFusion.cpp

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    projectedPoint = cPoint - projectionACToNormal;
+
+    //determine the equation of the line of the slope and sub in projected values (symmetric form)
+    double altitude = (((projectedPoint.x - aPoint.x) / vectorAI.x) + aPoint.z) * vectorAI.z;
+
+    return altitude;
+}
+
+double LandingManager::approachSpeed(const bool *ifPackage)
```
Fberbari: noted before, but emphasizing here. Reason to pass by reference is to save copying space/time. THat's good for custom structs, but since a bool and a pointer to a bool take up the same space, there's no need to pass a bool by reference.

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,202 @@
...
+    {
+        return STALL_SPEED_WITH_PACKAGE + 2;
+    }
+    else 
+    {
+        return STALL_SPEED_NO_PACKAGE + 2;
+    }
+}
+
+_LandingPath LandingManager::createSlopeWaypoints(const Telemetry_PIGO_t * input, const float * currentAltitude)
```
Fberbari: THere is a c++y way of passing these variables by reference. What you've got here is the old C way, which I started building into our code cuz I didn't know any better. Here's the right way to do it in C++.
https://www.ibm.com/docs/en/zos/2.2.0?topic=calls-pass-by-reference-c-only

The advantage to this is you don't ever have to worry about pointer de referencing or people passing in trash pointers ever again.

Autopilot/PathManager/Src/pathStateClasses.cpp
Fberbari: Ok I see. Really it's the landingTakeoff manager, no problem. I think changing the name to that would make things a bit clearer. THoughts ?

Autopilot/Src/SensorFusion.cpp
```diff
@@ -6,7 +6,13 @@
 #include "MadgwickAHRS.h"
 #include <math.h>
 
-SFError_t SF_GetResult(SFOutput_t *Output, IMU_Data_t *imudata, Airspeed_Data_t *airspeeddata) {
+typedef struct {
+    float roll, pitch, yaw; //rad
+    float rollRate, pitchRate, yawRate; //rad/s
+    float Airspeed; //m/s
```
Fberbari, line 12: track was added into here some time ago. THis is rebasing problem shit. Uh Oh. We're gonna have a hell of a time merging this LOOOL but we'll cross that bridge when we come to it.

Don't bother doing anything about it now though.

Fberbari (changes requested):
Ok, done. I think we're almost good to merge. I didn't detect any major problems here, so once you fix up the minor stuff I noted, I'll approve.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: oh yeah ok, that makes sense. Will fix

Autopilot/PathManager/Inc/trackFinder.hpp
NixonCTChan: I don't think sensorFusion has those changes yet though?

NixonCTChan:
- landing and takeoff finished!

Fberbari:
Closing cuz these changes went into another, cleaner PR that's now merged. This PR faced too many merged conflicts.

PR #109, opened 2021-04-03
IMU Header

Pop0097:
As requested by @Fberbari

### DIFF
--- a/Autopilot/Inc/IMU.hpp
+++ b/Autopilot/Inc/IMU.hpp
@@ -1,10 +1,16 @@
 /**
  * IMU Sensor Functions and Part Number Selection
- * Author: Lucy Gong
+ * Author(s): Lucy Gong, Dhruv Rawat
  */
 
 #ifndef IMU_HPP
 #define IMU_HPP
+
+#include <stdlib.h>
+#include <ctime>
+#include <cstdint>
+#include <math.h>
+#include "main.h"
 
 #define ICM_20602 0
 #define MPU9255 1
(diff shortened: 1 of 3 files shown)


Autopilot/Inc/IMU.hpp
```diff
@@ -1,28 +1,34 @@
...
 
-    bool isDataNew; 
+    bool isDataNew;
     int sensorStatus; //TBD but probably 0 = SUCCESS, -1 = FAIL, 1 = BUSY 
-    float utcTime; //Last time GetResult was called
+    uint32_t utcTime; //Last time GetResult was called
 };
 
-class IMU{
+class IMU {
```
Fberbari, line 31: Ah yes lmao

Autopilot/Inc/IMU.hpp
```diff
@@ -37,33 +43,45 @@ class IMU{
...
-
 //To be replaced with implementation of actual sensor
 class ICM20602: public IMU{
     public:
-        /**
-         * Initializes IMU
-         * */
-        void Init(){}; 
+        ICM20602(const ICM20602*) = delete;
+        static ICM20602* GetInstance();
```
Fberbari, line 50: THere's a much more c++y way of doing this it turns out. if we create a static object and return a reference to it,then we need not think about pointers. Don't do this now though. I'd like to change this eventually, but we'll change it for all the sensors one shot before their driver implementations are merged.

Fberbari (approved):
Good stuff let's merge!

PR #110, opened 2021-04-05
With this change, the old simulation results are removed when a new s…

Fberbari:
…im is run. This was always the intent, but the previous implementation didn't work reliably.


NixonCTChan (approved):
Looks good!

PR #112, opened 2021-04-13
Attitude manager now interfaces with new Sensor fusion design

Fberbari:
With this commit, attitude manager no longer deals with any hardware at
all. Instead, as per the new design of Sensor fusion, it just calls the
appropriate getter function from the sensor fusion module to obtain the
processed sensor data. The unit tests have also been updated to reflect
this fix.


NixonCTChan (approved):
Everything looks good!

Pop0097 (commented):
Looks good to me. But you may wanna delete the fetchSensorMeasurement files

Pop0097 (changes requested):
One thing I forgot. Since attitude manager is no longer accessing the IMU and Airpseed data directly, you can delete the `IMU_Data_t` and `Airspeed_Data_t` structs in `AttitudeDatatypes.hpp`.

We will instead redeclare those structs in `SensorFusion.hpp`.

Pop0097 (commented):
Thanks for deleting the files. By the looks of it the CMake needs to be modified though.

Some builds are failing because IMU_Data_t cannot be found and all. This will be addressed in Tony's PR


Screen Shot 2021-04-15 at 6 48 39 PM

Pop0097 (approved):
Looks good

PR #115, opened 2021-04-28
Final landing takeoff

NixonCTChan:
Brings over all the changes from the previous landing takeoff PR's, hopefully this will be able to merge easily.


Autopilot/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -138,4 +171,4 @@ attitudeState& FatalFailureMode::getInstance()
 {
     static FatalFailureMode singleton;
     return singleton;
-}
+}
```
Pop0097: newline

Autopilot/CMakeLists.txt
```diff
@@ -293,6 +298,7 @@ elseif(${KIND_OF_BUILD} STREQUAL "SIMULATION")
     ${CMAKE_CURRENT_SOURCE_DIR}/AttitudeManager/Src/OutputMixing.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/AttitudeManager/Src/SensorFusion.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/AttitudeManager/Src/MadgwickAHRS.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/AttitudeManager/Src/fetchSensorMeasurementsMode.cpp
```
Pop0097: I don't think this file exists anymore. Chief got rid of it in his last PR.

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -120,4 +124,13 @@ Gps_Data_t SF_GetRawGPS();
...
 Altimeter_Data_t SF_GetRawAltimeter();
 
+/**
+    The goal of this function is to return the current track of the plane in degrees, given its velocities in the latitude and longitude direction in m/s.
+
+    * @param[in] latitudeSpeed -> load in the speed of the plane in the N/S direction (change in latitude, m/s)
+    * @param[in] longitudeSpeed -> load in the speed of the plane in the E/W direction (change in longitude, m/s)
+    * @return double - > returns the direction of the resulting velocity/track of the plane
+*/
+double getTrack(double latitudeSpeed, double longitudeSpeed);
```
Fberbari: Na, not like this. track is just one of the fields of SFOutput, you'l get it from that structure. The user should never have to call this function. It will just become part of the inside mechanisms of sensor fusion.

To use track, just collect the SFOutput struct from the sensorFusion state and access the track field

Autopilot/CMakeLists.txt
Pop0097: Yeah, just checked, it isn't in your branch

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
+#pragma once
+
+/* Related Confluence Pages
+    https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1607630921/Landing+and+Takeoff
+*/
+
+#include "waypointManager.hpp"
+#include "vectorClass.hpp"
```
Fberbari: vector class does not seem to get used in this header file, right ? If that's the case then don't include it in the header in an effort to reduce dependancies. Include it in the cpp instead.

Is the same true with AutoSteer ?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -9,6 +9,9 @@
 //Frequency of SF calculations in Hz
 constexpr int SF_FREQ = 512;
 
+//Degree to radian conversion
+constexpr double DEG2RAD {3.1415926535};
```
Pop0097: I would rather you just use PI here since it is more clear :)). 

Also, in SensorFusion.cpp you include <math.h> which contains M_PI. If that doesn't work, you can do what is done in waypointManager.hpp with the #ifndef M_PI and all

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
...
+{
+    Telemetry_PIGO_t * telemetryData;
+    SFOutput_t * sensorOutput;
+};
+
+struct _LandingTakeoffOutput //this structure is used to hold the outputs of each landing state
+{
+    double desiredHeading; //desired heading is used when useHeading is true
+    double desiredTrack; //desired track is used when useHEading is false
+    double useHeading; //this variable is used to determine if heading needs to be controlled, rather than track
```
Fberbari: Is this meant to be a boolean ?

Autopilot/Inc/SensorFusion.hpp
```diff
@@ -120,4 +124,13 @@ Gps_Data_t SF_GetRawGPS();
  */ 
 Altimeter_Data_t SF_GetRawAltimeter();
 
+/**
+    The goal of this function is to return the current track of the plane in degrees, given its velocities in the latitude and longitude direction in m/s.
+
+    * @param[in] latitudeSpeed -> load in the speed of the plane in the N/S direction (change in latitude, m/s)
+    * @param[in] longitudeSpeed -> load in the speed of the plane in the E/W direction (change in longitude, m/s)
+    * @return double - > returns the direction of the resulting velocity/track of the plane
```
Pop0097: ->

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
...
+            This function translates landing takeoff data to output data
+            Takes in a landingTakeoffOutput instance, translates its information into the 2 types of coordinated turn inputs
+
+            @param _LandingTakeoffOutput* outputData - this pointer points to the output of the landing and takeoff states
+            @param SFOutput_t* sensorOutput - this pointer points to the sensor output of sensorFusion
+            @param IMUOutput_t* imuOutput - this pointer points to the imu output of sensorFusion
+            @param CoordinatedTurnInput_t * turnInput - this pointer points to the coorinated turn input structure that needs to be changed
+            @param AltitudeAirspeedInput_t * altitudeAirspeedInput - this pointer points to the altitude and airspeed input structure that needs to be changed
+        */
+        static void translateLTSFCommandsToCoordTurns(const _LandingTakeoffOutput & outputData, const SFOutput_t & sensorOutput, const IMU_Data_t & imuOutput, CoordinatedTurnInput_t & turnInput, AltitudeAirspeedInput_t & altitudeAirspeedInput);
```
Fberbari: This function and the one above are common to both takeoff and landing, right ? Would it be wise to add another comment seperation section to indicate that it's common ?

Autopilot/PathManager/Inc/waypointManager.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include <cstdint>
 #include <stdlib.h>
+#include "pathDatatypes.hpp"
```
Pop0097: If waypoint manager isn't using anything from this file, then no need to include it. 

A way to test for it is by commenting out the include statement and building again lol

Autopilot/PathManager/Inc/pathManager.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 #include "pathManagerStateManager.hpp"
+#include "landingTakeoffManager.hpp"
```
Fberbari: I think this file is not used in the header, right ? If that's the case, just stick it in the appropriate .cpp instead

Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -40,6 +40,36 @@ struct Telemetry_PIGO_t {
 
     Telemetry_Waypoint_Data_t waypoints[100]; // Somehow need to get PATH_BUFFER_SIZE here...
     Telemetry_Waypoint_Data_t homebase;
+
+    // landing and takeoff 
+    bool beginLanding;
+    bool beginTakeoff;
+    float stoppingDirectionHeading;
+    float stoppingLongitude;
+    float stoppingLatitude;
```
Pop0097: Should be doubles

Autopilot/PathManager/Inc/waypointManager.hpp
Fberbari: This doesen't look like it belongs here. Why did you decide to add it ?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
...
+    https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1607630921/Landing+and+Takeoff
+*/
+
+#include "waypointManager.hpp"
+#include "vectorClass.hpp"
+#include "AutoSteer.hpp"
+#include "SensorFusion.hpp"
+#include "pathDatatypes.hpp"
+
+enum _FlightStage{ROLL=0, CLIMB, CRUISING, TRANSITION, SLOPE, FLARE, DECRAB, TOUCHDOWN};
```
Pop0097: Do you think it would be best to put this enum in pathDatatypes.hpp? That way if anything needs it like pathManager.hpp it doesn't need to access the entire functionality of our landing system

Autopilot/PathManager/Inc/pathManager.hpp
Pop0097: Refer to comment in landingTakeoffManager.hpp

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+/**********************
+    TAKEOFF CONSTANTS
+**********************/
+constexpr static double EXIT_TAKEOFF_ALTITUDE {50}; //in meters, difference in altitude
+constexpr static double ROTATION_SPEED_NO_PACKAGE {4}; //in meters per second
+constexpr static double ROTATION_SPEED_WITH_PACKAGE {6}; //in meters per second
+constexpr static double CLIMB_SPEED_NO_PACKAGE {5}; // in meters per second
+constexpr static double CLIMB_SPEED_WITH_PACKAGE {7}; // in meters per second
+constexpr static double DISTANCE_OF_TAKEOFF {100}; // in meters, just used to approximate the horizontal location
```
Pop0097, line 35: Love that you put all the constants at the top!

Autopilot/PathManager/Src/waypointManager.cpp
```diff
@@ -91,8 +91,13 @@ _WaypointStatus WaypointManager::initialize_flight_path(_PathData ** initialWayp
     numWaypoints = numberOfWaypoints;
     nextFilledIndex = 0;
     
-    #ifdef UNIT_TESTING
-        currentIndex = 2;
+    #ifdef UNIT_TESTING 
+    //waypoint manager tests use a currentIndex of 2. To ensure that the tests work for flight paths with fewer waypoints, this if statement is employed
```
Fberbari, line 95: @Pop0097 this make sense to you ?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+struct _LandingPath //this structure is used to hold the waypoint dat of the landing path, when it is created in the transition stage
```
Pop0097: typo: waypoint data

Autopilot/Src/SensorFusion.cpp
```diff
@@ -86,3 +86,16 @@ SFError_t SF_GetAttitude(SFAttitudeOutput_t *Output, IMU_Data_t *imudata, Airspe
     return SFError;
 }
 
+double getTrack(double latitudeSpeed, double longitudeSpeed)
```
Fberbari: Like I say, not like this. User should not have to call this. Track is just another datapoint bundled in the sensor fusion struct.

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -0,0 +1,170 @@
...
+            Takes the one of the horizontal coordinates, plug it into the equation of the slope, get an altitude
+            
+            @param SFOutput_t * input - this variable will hold the sensor fusion data which is used to calculate current position
+            @param _PathData * aimingPoint - this variable will hold the location of the aiming point
+            @param _PathData * intersectionPoint - this variable will hold the location of the intersectionPoint
+            @param _PathData * stoppingPoint - this variable will hold the location of the stoppingPoint
+
+            @return double - this function will return the desiredAltitude for the position of the relative points
+        */
+        static double changingAltitude(const SFOutput_t & input, const _PathData & aimingPoint, const _PathData & intersectionPoint, const _PathData & stoppingPoint);
```
Pop0097, line 81: I love that you are using reference parameters here! Using proper C++ conventions lesgooooo

Fberbari (changes requested):
I did this all as one off comments. My b. But yeah, a few things to address, lmk if you think I'm wrong on any of them.

Apologies on missing them on my previous review. Things are much clearer in this one once I was able to know exactly what your'e changing

Autopilot/PathManager/Inc/pathManager.hpp
```diff
@@ -16,6 +17,12 @@ class pathManager
         void setState(pathManagerState& newState);
         _Path_Manager_Cycle_Status getStatus() {return status;}
 
+        //used to determine the stage of the landing sequence
+        _FlightStage stage;
+        bool madeLandingPoints;
+        bool madeTakeoffPoints;
+        bool isError;
+        bool isPackage;
```
Pop0097, line 25: If you don't mind entertaining me, we could try making these protected. 

```
protected:

         _FlightStage stage;
        bool madeLandingPoints;
        bool madeTakeoffPoints;
        bool isError;
        bool isPackage;

```

This means that only pathManager objects and objects of child classes can access these variables. It adds some encapsulation while also making the variables visible to our states. 

Thoughts on this?

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -133,22 +241,69 @@ void coordinateTurnElevation::execute(pathManager* pathMgr)
     _WaypointManager_Data_Out * waypointOutput = cruisingState::GetOutputData(); // Get output data from waypoint manager
     
     CoordinatedTurnInput_t turnInput {};
-    turnInput.currentHeadingTrack = sensFusionOutput->track; // Gets track;
-    turnInput.desiredHeadingTrack = waypointOutput->desiredTrack;
-    turnInput.accY = imudata->accy;
+    AltitudeAirspeedInput_t altAirspeedInput {};                                   
+    _LandingTakeoffOutput * landingTakeoffOutput;
```
Pop0097: A good practice is to always initialize pointers to `nullptr` (Actually, this applies to all variables. I debugged waypoint manager for hours only to find I forgot to initialize an integer). That way they don't point to random garbage and cause undefined behaviour if you are not careful.

I will try to find other places where you didn't initialize them, but maybe doing a quick check on your own for spots like this will be great!

Pop0097: If you don't want to add the = sign. Just add `{}`. It triggers value initialization, which will set pointers to nullptr and other datatypes to their defaults (ex. integers to 0)

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -133,22 +241,69 @@ void coordinateTurnElevation::execute(pathManager* pathMgr)
...
+            altAirspeedInput.currentAltitude = sensFusionOutput->altitude;
+            altAirspeedInput.desiredAltitude = waypointOutput->desiredAltitude;
+            altAirspeedInput.currentAirspeed = sensFusionOutput->airspeed;
+            altAirspeedInput.desiredAirspeed = waypointOutput->desiredAirspeed; 
+            
+    }
+
+    if(pathMgr->stage!=CRUISING)
+    {
+        LandingTakeoffManager::translateLTSFCommandsToCoordTurns(*landingTakeoffOutput, *sensFusionOutput, *imudata, turnInput, altAirspeedInput);
```
Pop0097, line 299: Happy noises. I see reference parameters

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -175,7 +330,360 @@ pathManagerState& fatalFailureMode::getInstance()
...
+
+    //max throttle
+    output.controlDetails.throttlePassby = true;
+    output.controlDetails.throttlePercent = 100;
+
+    if(!pathMgr->madeTakeoffPoints)
+    {
+        takeoffPoint = LandingTakeoffManager::createTakeoffWaypoint(input.sensorOutput->latitude,input.sensorOutput->longitude, input.sensorOutput->altitude, input.telemetryData->takeoffDirectionHeading);
+        pathArray[0] = takeoffPath.initialize_waypoint(takeoffPoint.longitude, takeoffPoint.latitude, takeoffPoint.altitude, PATH_FOLLOW);
+        currentLocation = takeoffPath.initialize_waypoint(input.sensorOutput->longitude, input.sensorOutput->latitude, input.sensorOutput->altitude, HOLD_WAYPOINT);
```
Pop0097: I'd recommend adding 10-15 metres in the altitude of currentLocation so it's not at ground level

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -0,0 +1,202 @@
+#include "landingTakeoffManager.hpp"
+#include <math.h>
```
Pop0097, line 3: If you defined PI in landingTakeoffManager.hpp, you can do:

```
#ifndef M_PI
#define M_PI 3.14159...
#endif
```

Pop0097 (commented):
Nice work! My comments are mostly nit-picky!

Autopilot/PathManager/Src/waypointManager.cpp
Pop0097: Yep. We had agreed on it

Autopilot/Inc/SensorFusion.hpp
NixonCTChan: so I don't even need this function? I thought we agreed sensorFusion would implement this in the future?

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
NixonCTChan: not for autosteer, the coordturns translation command needs the struct

Autopilot/PathManager/Inc/pathManager.hpp
NixonCTChan: the _FlightStage struct is used here, but i'm probably gonna move it to pathDatatypes like dhruv said.

Autopilot/PathManager/Inc/waypointManager.hpp
NixonCTChan: yup, doesn't belong. Will fix

Autopilot/Inc/SensorFusion.hpp
NixonCTChan: probably gonna delete this soon, seems like track is gonna be done by sensorfusion and it won't need a function to do that

Autopilot/PathManager/Inc/waypointManager.hpp
NixonCTChan: yup fixed

Autopilot/PathManager/Inc/pathDatatypes.hpp
NixonCTChan: true, fixed!

Autopilot/PathManager/Inc/pathManager.hpp
NixonCTChan: does that mean these variables can't be changed by the states? That's something we need, not just visibility.

Autopilot/PathManager/Src/landingTakeoffManager.cpp
NixonCTChan: what problem is this solving?

Pop0097 (commented):
Looks good to me. Will let @Fberbari take a last look before it's approved!

Autopilot/Inc/SensorFusion.hpp
Fberbari: Yes we did. But this function won't get exposed to the public. It will live inside the guts of sensorFusion with all of Tony's other stuff. When SF_GenerateNewResult is called by Freertos, one of the things that will happen is trackFinding, which will be done by your code. 

But that's Opaque to the user. All they see is a track value that gets set inside the result struct. 

It's a way of abstracting things so users don't have to worry about the complex stuff that goes on in SensorFusion (and it allows us to change SensorFusion guts in the future without having to modify the calling code and such).

Fberbari (approved):
I'm satisfied, and I'm approving this!!!!!! 

Wait on Dhruv's final go ahead but after that, merge merge merge!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #116, opened 2021-05-03
Telem and path manager comms

Gfcello:
@Fberbari Lmk if this is what you need.


Fberbari (approved):
all pretty textbook. Just needed so simulation work could be completed. Waiting for a passing CI and we'll be ready to merge.

PR #117, opened 2021-05-05
Simulation now involves Path manager as well as attitude manager

Fberbari:
Reason this PR touched so many files is because hooking in the Simulation stuff involved poking at a lot of stuff. But here are the big points:

The new mechanism for the simulation user to inject data is in AutoPilot/Simulation/Intercepts CommsWithTelemetry_intercept_user.cpp. There, a bunch of PIGO structs can be instantiated as you wish, and each time path manager executes a cycle, it will pick one up. 

Now, rather than the simulation data being injected out of the sensor interfaces and going through sensorFusion, the simulation data is injected right out of the sensor fusion interface and into the managers. This way, Simulation can focus on testing the core control algorithms while sensor fusion can be tested independently. 

Simulation main is in Autopilot\Simulation\SimDriver\simMain.cpp. That's where the managers are called. Atm attitude Manager is called the same number of times as path manager. If either manager ever enters failed state, a message is displayed and the simulation stops.

Comms between path and attitude managers are done by Autopilot\Simulation\Intercepts\CommsBetweenStateMachines_Intercept.cpp which doesn't involved FREERTOS. As a bonus, every time a communication occurs, the contents of the passed struct are dumped to a file in AutoPilot/SimulationBuild/DebugLogs.

I also had to move a couple things around. Namely, the PIGO struct definition had to be moved from PathDataTypes to TelemPathInterface.hpp. Otherwise, Telemetry manager wouldn't be able to access it. This means I had to add a #include "TelemPathInterface.hpp" to the cruising and landingTakeoff header files.

I gave a namespace to the enums for attitude and path manager statuses because their definitions were interfering with each other.

I had to modify the names of the mail queues between attitude and path and between path and telemetry for the same reason, they were interfering with each other.


Fberbari:
I gotta fix a couple failing unit tests, and go over the changes one more time myself before merging, but I'm ready to take a couple reviews.

Autopilot/CMakeLists.txt
```diff
@@ -272,43 +272,63 @@ elseif(${KIND_OF_BUILD} STREQUAL "SIMULATION")
...
     ${SIM_ROOT_DIR}/GeneratedCCode/SimulinkModel/Infinion_grt_rtw
-    ${SIM_ROOT_DIR}/GeneratedCCode/R2020a/extern/include
-    ${SIM_ROOT_DIR}/GeneratedCCode/R2020a/rtw/c/src
-    ${SIM_ROOT_DIR}/GeneratedCCode/R2020a/rtw/c/src/ext_mode/common
-    ${SIM_ROOT_DIR}/GeneratedCCode/R2020a/simulink/include
+    ${SIM_ROOT_DIR}/GeneratedCCode/Headers/extern/include
+    ${SIM_ROOT_DIR}/GeneratedCCode/Headers/rtw/c/src
+    ${SIM_ROOT_DIR}/GeneratedCCode/Headers/rtw/c/src/ext_mode/common
+    ${SIM_ROOT_DIR}/GeneratedCCode/Headers/simulink/include
   )
```
NixonCTChan, line 291: So from now on, we're gonna extract the generated code into this "Headers" folder?

NixonCTChan (commented):
Everything seems good. What was the reason for sensors outputting massive values? I couldn't find any changes that directly fixed that. Maybe in another PR?

Autopilot/CMakeLists.txt
Fberbari: Yeah, people who generate code will be required to rename this folder to Headers. Otherwise Cmake will break every quarter. I've got some code in the cmake that detects if this folder isn't renamed and instructs people to rename it if that's the case.

Fberbari:
> 
> 
> Everything seems good. What was the reason for sensors outputting massive values? I couldn't find any changes that directly fixed that. Maybe in another PR?

@NixonCTChan I'm having trouble remembering this issue. If it was a Matlab problem, hat would be in the other repo. Otherwise, it may have been a SensorFusion issue or attman issue, but I'm not sure. There's a bunch of changes here, so let's see if we can reproduce it, I'll fix it at that point.

NixonCTChan:
It was something with pitch not being restricted to a certain range. This prevents certain gains from being fed into the PID, since the pitch angle would be too high for the sim to run. I think it's probably an issue with sensor fusion, but here's a picture for reference. We'll see if we can reproduce it with these changes.

image

Autopilot/Test/Src/AttitudeManager/Test_Fsm.cpp
```diff
@@ -123,12 +123,9 @@ TEST(AttitudeManagerFSM, IfFetchInstructionsSucceedsTransitionToSensorFusion) {
 
 	attitudeManager attMng;
 
-	PMError_t PmNoError;
-	PmNoError.errorCode = 0;
-
 	/********************DEPENDENCIES*******************/
 
-	PM_GetCommands_fake.return_val = PmNoError;
+	GetCommands_fake.return_val = true;
```
Pop0097, line 128: I must be dumb, but did we declare GetCommands_fake anywhere?

Autopilot/PathManager/Inc/cruisingState.hpp
```diff
@@ -4,6 +4,7 @@
 */
 
 #include "pathDatatypes.hpp"
```
Pop0097: No need to include this anymore. All required structs are in `TelemPathInterface.hpp` now

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -657,9 +654,9 @@ void takeoffClimbStage::execute(pathManager* pathMgr)
 
         takeoffRollStage::waypointStatus = takeoffRollStage::takeoffPath.get_next_directions(waypointInput, &waypointOutput);
         output = LandingTakeoffManager::translateWaypointCommands(waypointOutput);
-
```
Pop0097, line 660: Did you go through all of these files with a formatter?

Autopilot/Simulation/Intercepts/CommsBetweenStateMachines_Intercept.cpp
```diff
@@ -0,0 +1,67 @@
...
+static void writeAttitudeCommandsToFile(CommandsForAM *commands);
+
+/***********************************************************************************************************************
+ * Code
+ **********************************************************************************************************************/
+
+void CommWithAMInit()
+{
+    PathToAttComLog.open("DebugLogs/PathToAttComLog.txt", std::fstream::out);
+    PathToAttComLog << "";
```
Pop0097, line 28: Maybe write "Initializing comms"?

Autopilot/Simulation/Intercepts/CommsWithTelemetry_intercept_user.cpp
```diff
@@ -0,0 +1,31 @@
...
+* Code
+********************************************************************/
+
+void CommWithTelemInit()
+{
+}
+
+void SendPathData(POGI *data)
+{
+ (void) data;
```
Pop0097, line 19: How does this compile?

Autopilot/Simulation/Intercepts/CommsBetweenStateMachines_Intercept.cpp
```diff
@@ -0,0 +1,67 @@
...
+void CommWithPMInit()
+{
+}
+
+void SendAttitudeData(AttitudeData *data)
+{
+    attitudeData = *data;
+}
+
+bool GetCommands(CommandsForAM *commands)
```
Pop0097, line 54: `GetPMData()`?

Pop0097 (commented):
Looks good to me. Small suggestions on my end.

Autopilot/Test/Src/AttitudeManager/Test_Fsm.cpp
Fberbari: this is actually how we do C style function mocking. the GetCommands function is ours, but it's implementation uses freeRtos. In our unit test build system, we've got fff (https://github.com/meekrosoft/fff). 

At the top of this file, you'l notice 
image

That macro tells fff to create a fake function that gets called instead of the real implementation. Along with it, all these definitions get created to allow us to fake its behaviour. GetCommands_fake instructions fff to return this value next time the code calls this function.

Autopilot/PathManager/Src/pathStateClasses.cpp
Fberbari: I screwed up.... My text editor by default removes all trailing white spaces. I turn it off for ZP dev but I forgot for this lol. Time to make a coding standard eih.

Autopilot/Simulation/Intercepts/CommsWithTelemetry_intercept_user.cpp
Fberbari: why wouldn't it ? (void) <insert arg here> is just a way to tell the compiler that we're not using that arg. It avoids us getting a warning.

Autopilot/Simulation/Intercepts/CommsBetweenStateMachines_Intercept.cpp
Fberbari: Do you mean to the terminal ?

Autopilot/Simulation/Intercepts/CommsBetweenStateMachines_Intercept.cpp
Fberbari: Yeah... I hate the naming too. but to change the name in the intercept, I have to change the name in the actual implementation file, which is already in devel. I didn't feel that renaming the files belonged in this PR. I'll create an asana entry for it though

Pop0097 (approved):
Looks good to me

Autopilot/Simulation/Intercepts/CommsWithTelemetry_intercept_user.cpp
Pop0097: Ah fair enough lol. I should've phrased that question nicer. Sorry

Autopilot/Simulation/Intercepts/CommsBetweenStateMachines_Intercept.cpp
Pop0097: Yeah, to the same logging file/terminal that the rest of the stuff is being logged to

Fberbari: Ah I see. this particular stuff gets logged to the DebugLogs/PathToAttComLog.txt file and that gets cleared on every run, so not too much use in writing the initialize thing.

PR #118, opened 2021-05-10
Added global math constants definition and fixed all custom implementations

Fberbari:
Turns out the C++ 11 standard cmath library does not include pi or e or the deg_to_rad functions. the _USE_MATH_DEFINES macro is a compiler specific thing. And none of our compilers seem to have it. Best not to use it for that reason anyway. Fine. Following the overwhelming recommendations of multiple stack overflow threads, I've just created a header file up in Common that contains a series of useful constants and functions that C++ doesn't give us. Anyone can include this file from anywhere in the Zero-Pilot project.

This PR also changes all current custom implementations of pi to rely on the global ones instead. Note that C++20 actually provides these things, but moving to C++20 is a large build system project in of itself.


Pop0097 (approved):
Looks good to me.

PR #119, opened 2021-05-13
Attitude/Path Manager Interface Fixes

antimonyB:
Rearranged where communication functions and structs are defined so they can be linked correctly and updated their names.


Fberbari:
A global description for what your PR has changed/what it achieves would be really good, and is something that should always be done

Autopilot/Inc/AttitudePathInterface.hpp
```diff
@@ -36,15 +36,6 @@ extern "C"
 
 const char PATH_ATTITUDE_MAIL_Q_SIZE = 1;
 
-//Set up a mail queue for sending commands to the attitude manager
-extern osMailQDef(commandsMailQ, PATH_ATTITUDE_MAIL_Q_SIZE, CommandsForAM);
```
Fberbari, line 40: I'm confused as to why your'e getting rid of this. How will each thread know about the other queue ?

Autopilot/PathManager/Src/CommWithAttitudeManager.cpp
```diff
@@ -4,18 +4,23 @@
...
 
 extern "C"
 {
 #include "cmsis_os.h"
 }
 
-void CommWithAMInit()
+//Set up a mail queue for sending commands to the attitude manager
+osMailQDef(commandsMailQ, PATH_ATTITUDE_MAIL_Q_SIZE, CommandsForAM);
+osMailQId commandsMailQ;
```
Fberbari: the name of the queue and the id are the same? Is this on purpose ? It's a bit confusing.

Autopilot/PathManager/Src/CommWithAttitudeManager.cpp
```diff
@@ -32,27 +37,28 @@ void SendCommandsForAM(CommandsForAM *commands)
...
     if(event.status == osEventMail)
     {
-        dataIn = static_cast<AttitudeData *>(event.value.p);
+        commandsIn = static_cast<CommandsForAM *>(event.value.p);
         
-        //Keep the data and remove it from the queue
-        *data = *dataIn;
-        osMailFree(attitudeDataMailQ, dataIn);
+        //Keep the command and remove it from the queue
+        *commands = *commandsIn;
```
Fberbari: I think something's wrong here. To me it looks like your'e reading and writing to/from the same queue, rather than actually exchange data with the other task.

Autopilot/PathManager/Src/CommWithAttitudeManager.cpp
```diff
@@ -32,27 +37,28 @@ void SendCommandsForAM(CommandsForAM *commands)
...
-bool GetAttitudeData(AttitudeData *data)
+
+bool GetFromPMToAM(CommandsForAM *commands)
 {
-    //Try to get data from mail queue
+    //Try to get commands from mail queue
     osEvent event;
-    AttitudeData * dataIn;
-    event = osMailGet(attitudeDataMailQ, 0);
+    CommandsForAM * commandsIn;
```
Fberbari: Your'e both reaing and writing commandsForAm. That's an issue. You should be writing commsWithAM, so that attitude manager can collect adn reading AttitudeData, so path manager can get attitude manager's data.

Fberbari (commented):
I can't see how this could work.

In your hardware tests, did you make sure you were able to write some data from one task, then read it from another task ?

Autopilot/Inc/AttitudePathInterface.hpp
antimonyB: These definitions still exist but have been moved to the cpp files to avoid linking issues.

Autopilot/PathManager/Src/CommWithAttitudeManager.cpp
antimonyB: I could rename the second one to commandsMailQID but I find this creates confusion on when to use commandsMailQ and when to use commandsMailQID as opposed to using a single name.

Autopilot/PathManager/Src/CommWithAttitudeManager.cpp
antimonyB: The header file still reads AttitudeData. Just the definitions have been moved so that each cpp file only deals with a single queue. Maybe this file should be renamed something like CommFromPMToAM.cpp as it no longer corresponds solely to CommWithAttitudeManager.hpp but rather provides some definitions for both header files.

Autopilot/PathManager/Src/CommWithAttitudeManager.cpp
Fberbari: I see what you mean, but to any code readers, this looks like you're using a variable before instantiating it. I would prefer using different names.

Fberbari:
Ok, I understand what's going on now, thx for the explanation. Pm and AM need to call functions from both files for this to work.

That's actually a really smart solution... no need to expose the queues to the world and share common files and stuff. Nice.

Only thing is it's really confusing, and you're right, I think it has to do with naming. Also, more explicit documentation in the Comments above the functions in the header files would clear a lot up. In this case, info about who needs to call what function.

Fberbari (approved):
Nice. This looks really good to me. Approved, Though give Dhruv a chance to review the PR before merging.

In the meantime, you should update whatever confluence docs exist on inter-state machine comms to reflect your new design.

Pop0097 (commented):
Looks good to me!

PR #121, opened 2021-05-24
GPS Driver (New and improved)

Pop0097:
This code is tested to work on my dev board. The sampling rate is around 1 Hz.

### DIFF
--- a/Autopilot/Inc/gps.hpp
+++ b/Autopilot/Inc/gps.hpp
@@ -8,10 +8,19 @@
 
 #include <stdint.h>
 
+/** What UART interface the device is connected to */
+#define GPS_UART_INTERFACE 1
+
+/** Baudrate the gps communicates over by default */
+#define GPS_UART_BAUDRATE 9600
+
+/** How large to make the internal buffer for parsing messages. Should be the size
+ of the largest message we'll receive from the gps*/
+#define GPS_UART_BUFFER_SIZE 800
+
 #define NEO_M8 0
 
-typedef struct
-{
+struct GpsData_t{
     long double latitude;  // 8 Bytes
     long double longitude; // 8 Bytes
     float utcTime;     // 4 Bytes. Time in seconds since 00:00 (midnight)
@@ -21,15 +30,15 @@ typedef struct
     uint8_t numSatellites;    // 1 Byte
 	uint8_t fixStatus; //0 = No GPS, 1 = GPS fix, 2 = DGSP Fix, 3 = Estimated/Dead Recoking Fix
 
-    uint8_t sensorStatus; // 0 = no fix, 1 = gps fix, 2 = differential gps fix (DGPS) (other codes are possible)
+    uint8_t sensorStatus; // 0 = good, 1 = error
     bool dataIsNew; // true if data has been refreshed since the previous time GetResult was called, false otherwise.
-	bool timeIsValid;
+	bool timeIsNew;
 
 	//Added these so autopilot knows which data is new
 	bool ggaDataIsNew; //Position, altitude, time, and number of satellites
 	bool vtgDataIsNew; //Groundspeed and Heading
 
-} GpsData_t;
+};
 
 class Gps
 {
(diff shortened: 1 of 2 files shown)


Autopilot/Inc/gps.hpp
```diff
@@ -49,51 +58,74 @@ class Gps
...
 	public:
 		NEOM8(const NEOM8*) = delete;
 		static NEOM8* GetInstance();
 
 		/**
-         * Triggers interrupt for new GPS measurement - stores raw data in variables and returns right away
-         * */
+		 * Triggers interrupt for new GPS measurement - stores raw data in variables and returns right away
+		 * */
 		void BeginMeasuring() {}; //No need to call this to get data. Just call GetResult()
```
Fberbari: you can actually just get rid of this. But a quick not somewhere that this gps just spews data without anyone asking would be good.

Autopilot/Inc/gps.hpp
```diff
@@ -49,51 +58,74 @@ class Gps
...
+		 * 1. Reset dataIsNew flag
+		 * 2. Transfers raw data from variables to struct
+		 * 3. Updates utcTime and status values in struct as well
+		 * */
 		void GetResult(GpsData_t *Data);
 
+		/**
+		 * Returns the buffer used to receive GPS UART signals
+		 */
+		uint8_t* get_byte_collection_buffer();
```
Fberbari, line 80: Do this func and the one below need to be public ? Or can they be private ?

Autopilot/Inc/gps.hpp
```diff
@@ -49,51 +58,74 @@ class Gps
...
-		uint8_t uint8_to_hex(unsigned int);
-		int ascii_to_hex(unsigned int);
-		void parse_gps_data();
-		void parse_vtg();
-		void parse_gga();
+		static NEOM8* gpsInstance;
+
+		// Buffers
+		uint8_t gga_buffer[GPS_UART_BUFFER_SIZE]; //buffer for pasing gga (positional packets)
+		uint8_t vtg_buffer[GPS_UART_BUFFER_SIZE]; //buffer for parsing vtg packets (velocity packets)
```
Fberbari, line 93: ahahaha 3.2KB of buffers. Nice. All good, but hillarious

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,450 @@
...
+#define GPS_GGA_MESSAGE "GNGGA"
+#define GPS_VTG_MESSAGE "GNVTG"
+
+extern UART_HandleTypeDef huart4;
+
+NEOM8* NEOM8::gpsInstance = nullptr;
+
+NEOM8* NEOM8::GetInstance() {
+	if (!gpsInstance) {
+		gpsInstance = new NEOM8 {};
```
Fberbari, line 71: if you'd like to, there's a more c++ way to do this. It's what we do in the state machines, here's an example:

attitudeState& sensorFusionMode::getInstance()
{
    static sensorFusionMode singleton;
    return singleton;
}

Just means we don't have to deal with pointers, only references.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+/**
+ * Method is called when our uart buffer (byte_collection_buffer) is full
+ */
+void HAL_UART_RxCpltCallback(UART_HandleTypeDef *huart) {
+	NEOM8 * neoM8N = NEOM8::GetInstance();
+
+	neoM8N->parse_gpsData();
+
+	HAL_UART_Receive_DMA(&huart4, neoM8N->get_byte_collection_buffer(), GPS_UART_BUFFER_SIZE);
```
Fberbari, line 113: Oh cool! You just ask again as soon as data comes in. Smart.

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,450 @@
...
+	HAL_UART_Receive_DMA(&huart4, byte_collection_buffer, GPS_UART_BUFFER_SIZE);
+}
+
+/**
+ * Method is called when our uart buffer (byte_collection_buffer) is full
+ */
+void HAL_UART_RxCpltCallback(UART_HandleTypeDef *huart) {
+	NEOM8 * neoM8N = NEOM8::GetInstance();
+
+	neoM8N->parse_gpsData();
```
Fberbari: So there's a rule about doing too many things in interrupt service routines. The rule is you should do as little as possible. Reason being that interrupts are kinda wild and put everything on pause. If some other interrupt (maybe the IMU one for instance) were to fire while this ISR is running, well the processor would have to put one of them on the back burner. That would potentially mean that while the processor is parsing the GPS data, it's ignoring data that the IMU is sending. 

So interrupts shouldn't ever do any data processing. Just get the data, store it and exit. You should get your getResult function to do the processing instead.

Fberbari (commented):
This is looking really good!!

Autopilot/Inc/gps.hpp
Pop0097: They need to be public for the interrupt unfortunately

Autopilot/Inc/gps.hpp
Pop0097: We can always download more RAM. Also, if we can reduce the message size through actually configuring the sensor then we can reduce the buffer size significantly

Autopilot/Src/gps.cpp
Pop0097: Looks neat!

Autopilot/Src/gps.cpp
Pop0097: Sounds good

Pop0097: But

if I parse outside, what happens if the data gets written over?

Fberbari: If we do things correctly, that should never happen. Sensor fusion should always call getData at least as fast as the gps can send us new data points. This is the benefit of having a real time OS. Otherwise, we'd need a FIFO

Pop0097: excellent point

Autopilot/Src/gps.cpp
```diff
@@ -0,0 +1,456 @@
...
+				 dataAvailable {false},
+				 uartDataIsNew {false} {
+
+//	const uint8_t CFG_NMEA[16] = { 0x17, 0x20, 0b00011000, 0x40, 0x08, 0x01, 0x00, 0x00, 0x00, 0b01110110, 0x00, 0x01, 0x01, 0x01, 0x00, 0x00 };
+//	HAL_UART_Transmit_DMA(&huart4, (uint8_t*) CFG_NMEA, sizeof(CFG_NMEA));
+//
+//	HAL_UART_Transmit_DMA(&huart4, (uint8_t*) PUBX_CONFIG_NMEA, sizeof(PUBX_CONFIG_NMEA));
+//
+//	//	const uint8_t CFG_UART[21] = { 0x06, 0x00, 0x01, 0x00, 0b00000000, 0b00000000, 0x00, 0x00, 0b11000000, 0b00001000, 0x00, 0x01, 0b11000010, 0x00, 0x00, 0b00000010, 0x00, 0b00000010, 0x00, 0x00, 0x00 };
+//	//	HAL_UART_Transmit_DMA(&huart4, CFG_UART, sizeof(CFG_UART));
```
Fberbari: Is all this stuff meant to be attempts at configuring the gps ? It's fine to keep it for the next guy, but walls of comments like this are terribly confusing. What you should do is surround the section with something like

#if 0
#endif

and just leave a line or two of comments above that explaining what this could be useful for.

Fberbari (approved):
One last minor comment. But otherwise, this is on track to being the first driver that gets merged into devel!

Autopilot/Src/gps.cpp
Pop0097: kk

PR #123, opened 2021-06-01
Telemetry framework Gordon

Gfcello:
Still waiting on a few more commits before this can be merged


Autopilot/Inc/TelemPathInterface.hpp
```diff
@@ -82,7 +82,6 @@ const char PATH_TELEM_MAIL_Q_SIZE = 1;
 extern osMailQDef(PMcommandsMailQ, PATH_TELEM_MAIL_Q_SIZE, Telemetry_PIGO_t);
 extern osMailQId PMcommandsMailQ;
 
-
```
Pop0097, line 85: Try to not include files where you only change whitespace in your PRs. It can lead to merge conflicts for others down the line...

Autopilot/Inc/xbee.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include "TelemPathInterface.hpp" //Holds structs for sent and received data
+
+//Ask Dhruv R (Elias) about UART stuff (Or ask our lord FREERTOS in the sky)
+
+class XBEE{
+    public:
+        /**
+         * Initializes XBEE
+         * */
+        virtual void Init() = 0;
```
Pop0097: For C++ classes you should use a constructor instead of an init() method. Since this is an ABC you should make the constructor `protected`

Also, is there a reason we are making this an abstract base class (ABC)? ABCs are really only used in inheritance, so if you are not inheriting it we may need to rethink how we want to structure this class.

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -1,12 +1,23 @@
 /*
 * Author: Nixon Chan
+* Implementing Telem Comms: Gordon Fountain
 */
 
 #include "telemetryStateClasses.hpp"
 
+
+Telemetry_PIGO_t _PMData;
+POGI GSData;
```
Pop0097: Let's rename some of these for consistency: `POGI --> Telemetry_POGI_t`, `GSData --> _GSData`

Also, these should be declared as `static` so they only have file scope and not global scope ;)

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -19,7 +30,9 @@ telemetryState& initialMode::getInstance()
 void obtainDataMode::execute(telemetryManager* telemetryMgr)
 {
     //obtain data from ground
+    //Receive_GS_Data(); //Receives data in MavLink form from the XBEE
```
Pop0097: I don't think this function call will work because the Xbee class is an ABC (and the method is a pure virtual method). 

Maybe we should consider making the class a singleton? We should set up a call about this. I'm free pretty much all weekend

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -186,6 +187,10 @@ telemetryState& encodeDataMode::getInstance()
 void sendDataMode::execute(telemetryManager* telemetryMgr)
 {
     //send data to ground
+    //Not sure if this state is needed. From what I understand FREERTOS is calling the function to send down data.
+    //Send_GS_Data(&_GSData);
```
Pop0097: @Fberbari

Pop0097: My understanding so far: this needs to stay as a state. FreeRTOS only calls the `.execute()` method for the Telemetry Manager, and the state is responsible for the instructions that are performed. This method will be called at a set period thanks to Dhruv U's work

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -95,11 +117,36 @@ telemetryState& readFromPathMode::getInstance()
...
+    telemetryMgr->dataValid &= &_PMData.numWaypoints != nullptr;
+    telemetryMgr->dataValid &= &_PMData.prevId != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingAltitude != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingDirectionHeading != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingLatitude != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingLongitude != nullptr;
+    telemetryMgr->dataValid &= &_PMData.takeoffDirectionHeading != nullptr;
+    telemetryMgr->dataValid &= &_PMData.waypointModifyFlightPathCommand != nullptr;
+    telemetryMgr->dataValid &= &_PMData.waypointNextDirectionsCommand != nullptr;
+    telemetryMgr->dataValid &= &_PMData.waypoints != nullptr;
```
Pop0097: Is there a reason we are analyzing the data (decoded by `obtainDataMode`) after sending it to pathmanager?

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -39,6 +52,9 @@ telemetryState& obtainDataMode::getInstance()
 void decodeDataMode::execute(telemetryManager* telemetryMgr)
 {
     //decode data with Mavlink
+    //Call Jingting's function here
+
+    //State change:
```
Pop0097: Another thing to add here: formatting the data so it fits in the struct (unless this is already done automatically)

Autopilot/TelemetryManager/CommsWithPathManager.cpp
```diff
@@ -0,0 +1,58 @@
+/**
+ * Implements methods for sending commands to the path manager from telemetry.
+ * Author: Anthony Bertnyk, messed with by Gordon Fountain to co-opt for Telemetry
+ */
+
+#include "CommsWithPathManager.hpp"
+
+extern "C"
+{
```
Pop0097: You have this file in your PR twice. Which one do we want to keep?

Pop0097: I'd delete this one since it looks like you worked on the other one more

Pop0097 (commented):
Good work so far!

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: Jinging's function should be sending the data out in the same struct, so I think we're good for that.

Autopilot/Inc/xbee.hpp
Gfcello: This was a very early design for it, I stopped working on it when I got blocked on the actual xbees, not soo sure how this will actually be implemented.

Gfcello: I would appreciate any suggestions you have

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: Okay, will also change POGI in the telem path interface

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: this is that data coming from the path manager, called after the readFromPathMode execute iirc

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: Sounds good

Autopilot/Inc/TelemPathInterface.hpp
```diff
@@ -25,7 +25,7 @@ struct Telemetry_Waypoint_Data_t {
 
 
 //Data for path manager to send to telemetry.
-typedef struct POGI{
+typedef struct Telemetry_POGI_t{
```
Pop0097: no need for the typedef:

```cpp
struct Telemetry_POGI_t {

};
```

If you do need the typedef cause of compiler issues (shouldn't happen), the syntax is this:

```cpp
typedef struct {

} Telemetry_POGI_t;
```

Pop0097:
What Gordon and I discussed today:

```txt
1. Make XBee class a singleton class, and have all classes that use it simply get an instance 
    1. Getting the instance can be done in the constructor of the states 
    2. For examples, refer to IMU driver 
        1. IMU.hpp
            1. getInstance() method (line 58)
            2. deleting constructor (line 63)
            3. Private constructor (line 70)
2. Add fake functions to your TelemetryFSM tests. Use Attitude and Path mananger FSM tests as examples
    1. Use Apple tests on CI to find which functions need to be mocked.
```

Autopilot/Inc/xbee.hpp
```diff
@@ -0,0 +1,44 @@
...
+        /*SendResult should:
+         * 1. Reset newOutput flag
+         * 2. Transfers raw data from variables to struct
+         * 3. Updates utcTime and status values in struct as well
+         * 4. Wait for FREERTOS to call the Send_Data function
+        */
+        static void SendResult(Telemetry_POGI_t* data);
+};
+
+#endif
```
Pop0097: Make into a singleton class as discussed

Autopilot/Test/Src/TelemetryManager/Test_TelemetryManager_Fsm.cpp
```diff
@@ -71,23 +71,13 @@ TEST(TelemetryManagerFSM, readToAnalyze){ //testing transition from readFromPath
     EXPECT_EQ(*(telemetryMng.getCurrentState()), analyzeDataMode::getInstance());
 }
 
-TEST(TelemetryManagerFSM, analyzeToReport){ //testing transition from analyzeDataMode to reportMode
+TEST(TelemetryManagerFSM, analyzeToEncode){ //testing transition from analyzeDataMode to reportMode
```
Pop0097: Add fake functions as discussed

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -19,7 +29,9 @@ telemetryState& initialMode::getInstance()
 void obtainDataMode::execute(telemetryManager* telemetryMgr)
 {
     //obtain data from ground
+    XBEE::Receive_GS_Data(); //Receives data in MavLink form from the XBEE
```
Pop0097: With singletons, this will not be needed anymore. Instead you will have the XBEE class insatnce as a parameter of the `obtainDataMode` class. You can initialize the parameter in the constructor of the `obtainDataMode` class in the `telemetryStateClasses.hpp` file by calling the getInstance() in XBEE method when you make it

 within the constructor of the `obtainDataMode` class you will call an ins

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -95,18 +116,43 @@ telemetryState& readFromPathMode::getInstance()
 
 void analyzeDataMode::execute(telemetryManager* telemetryMgr)
 {
-    //set the dataValid here
+    
+    telemetryMgr->dataValid = 1;
+    telemetryMgr->dataValid &= sizeof(&_PMData) == 19;
+    telemetryMgr->dataValid &= &_PMData.beginLanding != nullptr;
```
Pop0097: many of these fields are statically defined. Checking for `nullptr` will not work since they have a memory address. 

Instead you should check if their values fall within an acceptable range/set

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -95,18 +116,43 @@ telemetryState& readFromPathMode::getInstance()
...
+    telemetryMgr->dataValid &= &_PMData.numWaypoints != nullptr;
+    telemetryMgr->dataValid &= &_PMData.prevId != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingAltitude != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingDirectionHeading != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingLatitude != nullptr;
+    telemetryMgr->dataValid &= &_PMData.stoppingLongitude != nullptr;
+    telemetryMgr->dataValid &= &_PMData.takeoffDirectionHeading != nullptr;
+    telemetryMgr->dataValid &= &_PMData.waypointModifyFlightPathCommand != nullptr;
+    telemetryMgr->dataValid &= &_PMData.waypointNextDirectionsCommand != nullptr;
+    telemetryMgr->dataValid &= &_PMData.waypoints != nullptr;
```
Pop0097: _PMData.waypoints is a pointer and does not need the & sign since the variable stores the memory address.

Autopilot/TelemetryManager/Inc/telemetryStateClasses.hpp
```diff
@@ -19,6 +24,7 @@ class initialMode: public telemetryState
         initialMode(){}
         initialMode(const initialMode& other);
         initialMode& operator =(const initialMode& other);
+        void TelemCommWithPMInit();
```
Pop0097: You do not need to add `TelemCommWithPMInit()` here. In fact by doing this you will now call this undefined function (which is incorrect) whenever you write `TelemCommWithPMInit()`. 

Since you are importing `CommsWithPathManager.hpp` you will have access to this function and will just have to call `TelemCommWithPMInit()` to call the right one. 

Tl;dr: delete this

Autopilot/TelemetryManager/Inc/telemetryStateClasses.hpp
```diff
@@ -61,6 +67,7 @@ class passToPathMode: public telemetryState
         passToPathMode(){}
         passToPathMode(const passToPathMode& other);
         passToPathMode& operator =(const passToPathMode& other);
+        void SendCommandsForPM(Telemetry_PIGO_t _PMData);
```
Pop0097: same here

Autopilot/Inc/xbee.hpp
Gfcello: Implemented in 4ff2194

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: Pending ranges or sets

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: also fixed in 4ff2194

Autopilot/TelemetryManager/Inc/telemetryStateClasses.hpp
Gfcello: in d124597

Autopilot/TelemetryManager/Inc/telemetryStateClasses.hpp
Gfcello: in d124597

Autopilot/TelemetryManager/Inc/telemetryStateClasses.hpp
```diff
@@ -39,6 +38,7 @@ class obtainDataMode: public telemetryState
         obtainDataMode(){}
```
Pop0097: Should call the getinstance method for XBEE in this constructor

Pop0097 (commented):
Little comment. Let's hope travis works so you can get those tests to build

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -118,29 +118,39 @@ void analyzeDataMode::execute(telemetryManager* telemetryMgr)
...
+    telemetryMgr->dataValid &= _PMData.holdingTurnRadius > 10;
+    telemetryMgr->dataValid &= _PMData.initializingHomeBase == true || _PMData.initializingHomeBase == false;
+    if(_PMData.initializingHomeBase) {
+        telemetryMgr->dataValid &= (_PMData.homebase.waypointType == 0 || _PMData.homebase.waypointType == 1 || _PMData.homebase.waypointType == 2);
+    }
+    telemetryMgr->dataValid &= _PMData.modifyId > 0;
+    telemetryMgr->dataValid &= _PMData.nextId > 0;
+    telemetryMgr->dataValid &= _PMData.prevId > 0;
+    telemetryMgr->dataValid &= _PMData.stoppingAltitude >= 0;
+    telemetryMgr->dataValid &= _PMData.stoppingDirectionHeading >= 0;
```
Pop0097: forgot, this should be less than 360 too!

my apologies

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -118,29 +118,39 @@ void analyzeDataMode::execute(telemetryManager* telemetryMgr)
...
+    telemetryMgr->dataValid &= _PMData.stoppingLatitude >= 0;
+    telemetryMgr->dataValid &= _PMData.stoppingLongitude >= 0;
+    telemetryMgr->dataValid &= _PMData.takeoffDirectionHeading >= 0;
```
Pop0097: same here. less than 360 degrees too :))

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -118,29 +118,39 @@ void analyzeDataMode::execute(telemetryManager* telemetryMgr)
...
+    telemetryMgr->dataValid &= _PMData.waypointModifyFlightPathCommand > 0 && _PMData.waypointModifyFlightPathCommand <= 6;
+    telemetryMgr->dataValid &= _PMData.waypointNextDirectionsCommand > 0 && _PMData.waypointNextDirectionsCommand <= 2;
+    if((_PMData.waypoints == 0 && _PMData.waypoints != nullptr) || _PMData.waypoints < 0) {
```
Pop0097: you used waypoints instead of `numwaypoints` when comparing to an integer

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
Gfcello: altitude less than 360 too? Or just heading and the IDs?

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -130,13 +130,13 @@ void analyzeDataMode::execute(telemetryManager* telemetryMgr)
...
-    telemetryMgr->dataValid &= _PMData.stoppingDirectionHeading >= 0;
+    telemetryMgr->dataValid &= _PMData.stoppingDirectionHeading >= 0 && _PWData.stoppingDirectionHeading < 360;
     telemetryMgr->dataValid &= _PMData.stoppingLatitude >= 0;
     telemetryMgr->dataValid &= _PMData.stoppingLongitude >= 0;
-    telemetryMgr->dataValid &= _PMData.takeoffDirectionHeading >= 0;
+    telemetryMgr->dataValid &= _PMData.takeoffDirectionHeading >= 0 && _PWData.takeoffDirectionHeading < 360;
     telemetryMgr->dataValid &= _PMData.waypointModifyFlightPathCommand > 0 && _PMData.waypointModifyFlightPathCommand <= 6;
     telemetryMgr->dataValid &= _PMData.waypointNextDirectionsCommand > 0 && _PMData.waypointNextDirectionsCommand <= 2;
-    if((_PMData.waypoints == 0 && _PMData.waypoints != nullptr) || _PMData.waypoints < 0) {
+    if((_PMData.numWaypoints == 0 && _PMData.waypoints != nullptr) || _PMData.waypoints < 0) {
```
Pop0097: `if((_PMData.numWaypoints == 0 && _PMData.waypoints != nullptr) || _PMData.numWaypoints < 0) {`

Autopilot/TelemetryManager/Inc/telemetryStateClasses.hpp
```diff
@@ -33,6 +33,7 @@ class obtainDataMode: public telemetryState
         void execute(telemetryManager* telemetryMgr);
         void exit(telemetryManager* telemetryMgr){(void) telemetryMgr;}
         static telemetryState& getInstance();
+	static XBEE& getInstance();
```
Pop0097: You should not be making a method like this here. Since the XBEE class is a singleton, if a class wants to use an XBEE, it should just call the getInstance() method in teh XBEE class. 

If this is a result of confusion from my previous comment, this is what I meant :))

```cpp
class obtainDataMode: public telemetryState{    
public:        
  void enter(telemetryManager* telemetryMgr){(void) telemetryMgr;}       
  void execute(telemetryManager* telemetryMgr);       
  void exit(telemetryManager* telemetryMgr){(void) telemetryMgr;}        
  static telemetryState& getInstance();

private:        
  obtainDataMode(){ ZPXbee = XBEE::getInstance(); }        
  obtainDataMode(const obtainDataMode& other);        
  obtainDataMode& operator =(const obtainDataMode& other);        
  XBEE ZPXbee;
};
```

Autopilot/PathManager/Inc/CommsWithTelemetry.hpp
```diff
@@ -11,19 +11,19 @@
...
-void SendPathData(Telemetry_POGI_t *data);
+void SendFromPMToTM(Telemetry_POGI_t *commands);
 
 /**
  * Retrieve commands from the path manager.
  * @param[out]   commands   Pointer to struct to recieve commands.
  * @return True if commands were recieved.
  */
-bool GetTelemetryCommands(Telemetry_PIGO_t *commands);
+bool GetFromPMToTM(Telemetry_POGI_t *commands);
```
antimonyB, line 27: In the other interface, this is organized differently since this function is called by telemetry, not path manager. To be more consistent, this function declaration would be swapped with "GetFromTMToPM(Telemetry_PIGO_t *commands)", which is called by path manager. The current way works, but forces telemetry to include this header instead of keeping them isolated.

Autopilot/TelemetryManager/Inc/CommsWithPathManager.hpp
```diff
@@ -11,19 +11,19 @@
...
-void SendCommandsForPM(Telemetry_PIGO_t *commands);
+void SendFromTMToPM(Telemetry_PIGO_t *commands);
 
 /**
  * Retrieve data from telemetry.
  * @param[out]   data   Pointer to struct to recieve data.
  * @return True if data was recieved.
  */
-bool GetTelemData(Telemetry_POGI_t *data);
+bool GetFromTMToPM(Telemetry_PIGO_t *commands);
```
antimonyB, line 27: Again, this declaration could be swapped with GetFromPMToTM(Telemetry_POGI_t *commands) for consistency/organization.

Autopilot/TelemetryManager/Src/telemetryStateClasses.cpp
```diff
@@ -1,24 +1,34 @@
...
 {
-    //initial mode
-    //Initialize mail queue to send data to Path Manager:
-    TelemCommWithPMInit();
-    //Initializa mail queue to receive Path Manager data
+    //Initialize mail queue to SEND data to Path Manager
+    CommFromTMToPMInit();
+    
+    //Initialize mail queue to RECEIVE Path Manager data
+    CommFromPMToTMInit();
```
antimonyB: This is redundant since each module initializes its own outgoing mail queue before sending data. If it is not already initialized, that means the other module has not yet sent any data, so GetFromPMToTM would return false either way.

Autopilot/PathManager/Inc/pathStateClasses.hpp
```diff
@@ -45,11 +46,14 @@ class commsWithTelemetry : public pathManagerState
         void exit(pathManager* pathMgr) {(void) pathMgr;}
         static pathManagerState& getInstance();
         static Telemetry_PIGO_t* GetTelemetryIncomingData(void) {return &_incomingData;}
+        static Telemetry_POGI_t* GetTelemetryIncomingOutgoingData(void) {return &_outgoingData;}
+
     private:
-        commsWithTelemetry() {CommWithTelemInit();}
+        commsWithTelemetry() { CommFromTMToPMInit(); CommFromPMToTMInit(); }
```
antimonyB: CommFromTMToPMInit() is already called by telemetry and should only be called once.

Autopilot/Inc/xbee.hpp
```diff
@@ -0,0 +1,54 @@
...
+        /*Interrupt to send data to ground       
+        */
+        void Send_GS_Data(); //Only touched by FREERTOS
+
+        /*SendResult should:
+         * 1. Reset newOutput flag
+         * 2. Transfers raw data from variables to struct
+         * 3. Updates utcTime and status values in struct as well
+         * 4. Wait for FREERTOS to call the Send_Data function
+        */
```
Pop0097: Do all of these methods need to be public? I feel like Receive_GS_Data() and Send_GS_Data() can be private

Autopilot/Inc/xbee.hpp
```diff
@@ -0,0 +1,54 @@
...
+        void SendResult(uint8_t* data);
+    
+    private:
+        XBEE();
+
+        static XBEE* xbeeInstance;
+};
```
Pop0097: should have private `xbee_in_t` field to store outputs as they come

Autopilot/Src/xbee.cpp
```diff
@@ -0,0 +1,43 @@
+#include "xbee.hpp"
+#include "stm32f7xx_hal_usart.h"
+
+extern USART_HandleTypeDef husart2;
```
Pop0097: You need to set up the ZP ioc file so your I/O configs are seen here

Autopilot/Src/xbee.cpp
```diff
@@ -0,0 +1,43 @@
...
+	HAL_USART_Receive(&husart2, &xbeeIn.len, 1, 100);
+	uint8_t payload[xbeeIn.len];
+	for(int i = 2; i < xbeeIn.len; i++) {
+		HAL_USART_Receive_DMA(&husart2, &payload[i], 1);
+	}
+	xbeeIn.payload = payload;
+	return xbeeIn;
+}
+
+void XBEE::SendResult(Telemetry_POGI_t* data) {
```
Pop0097: parameter here is incorrect. Should be `uint8_t*`

Autopilot/Src/xbee.cpp
```diff
@@ -0,0 +1,43 @@
...
+	for(unsigned int i = 0; i < sizeof(data); i++) {
```
Pop0097: fun fact, `sizeof(data)` won't work. This is because of a phenomenon called pointer decay, where by passing a pointer array as a parameter you lose sight of the size of the array. As a result, sizeof(data) will return 1 byte, even if your array is 100 bytes long. You need a second parameter for the length of the array.

Autopilot/Src/xbee.cpp
```diff
@@ -0,0 +1,43 @@
...
+		HAL_USART_Transmit(&husart2, &data[i], 1, 100);
```
Pop0097: This is a blocking function, meaning that as it runs, literally every single operation on our board will be halted. Please consider using `HAL_USART_Transmit_IT` or `HAL_USART_Transmit_DMA` to overcome this.

This is a good video explaining it: https://www.youtube.com/watch?v=l5msGz2tZP0&t=2s

Autopilot/Src/xbee.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+void XBEE::Receive_GS_Data() {}
+
+void XBEE::Send_GS_Data() {}
+
+xbee_in_t GetResult() {
+	uint8_t holder;
+	xbee_in_t xbeeIn;
+	HAL_USART_Receive(&husart2, &holder, 1, 100);
+	HAL_USART_Receive(&husart2, &xbeeIn.len, 1, 100);
```
Pop0097: 1. Calling to the xbees should be done in `Receive_GS_Data()`. This method is a getter, so its role is to just send information out, not gather it
2. `HAL_USART_Receive` is blocking. refer to other comment. 

I don't think this will behave as you expect too... 

If this is anything like GPS, then the xbees will be constantly spitting data at you. This means at any given time you cannot be sure if you are getting the length or some other data. I feel like using DMA and a parser like I did with GPS would be your best bet. We can call to discuss more

@Fberbari you have any experience or say here?

Autopilot/Src/xbee.cpp
```diff
@@ -0,0 +1,43 @@
...
+	uint8_t payload[xbeeIn.len];
```
Pop0097: If an array has a variable length, you must allocate it on the heap. Additioanlly, you must make sure you clear this memory so we don't have a memory leak.

Autopilot/PathManager/Inc/CommsWithTelemetry.hpp
Pop0097: Oh okay. So I will just do an ol' switcheroo

Pop0097: Turns out switching them causes some odd linking errors. I'll keep them as is

Autopilot/Inc/xbee.hpp
Gfcello: Probably make more sense to be private, only were public for maybe FREERTOS stuff (No longer a thing).

Autopilot/Src/xbee.cpp
Gfcello: The goal is DMA, but DMA has not yet transmitted properly.

PR #124, opened 2021-06-05
Completed the IMU driver

Fberbari:
This driver is interrupt based and has been tested on hardware.


Autopilot/Inc/imu.hpp
```diff
@@ -0,0 +1,94 @@
...
+        void PrepareMagForDataMode(void);
+
+        void Bmx160WriteReg(uint8_t reg, uint8_t val);
+        void Bmx160ReadReg(uint8_t const regAddr, uint8_t *pData, uint8_t len);
+
+        //Variables
+        uint8_t rawImuData[21];
+
+};
+
```
Pop0097, line 86: May want to end the unit test and simulation stuff here. Similar to how we are doing it in other sensor drivers

Autopilot/Src/imu.cpp
```diff
@@ -0,0 +1,219 @@
...
+    Bmx160WriteReg(MAG_IF_2_REG, MAG_IF_2_DATA_MODE);
+    Bmx160WriteReg(MAG_IF_1_REG, MAG_IF_1_DATA_MODE);
+}
+
+void BMX160::Bmx160WriteReg(uint8_t reg, uint8_t val)
+{
+    uint8_t rx[2] = { 0, 0 }; // we don't care about what gets sent back to us in write mode
+    uint8_t tx[2] = { static_cast<uint8_t> (reg | BMX160_WRITE_BIT), val }; // Set first bit to 0
+
+    HAL_GPIO_WritePin(GPIOG, GPIO_PIN_10, GPIO_PIN_RESET);  // assert slave select
```
Pop0097: Do we need to change the value of this GPIO pin after sending and receiving our data?

Autopilot/Src/imu.cpp
```diff
@@ -0,0 +1,219 @@
...
+    Bmx160WriteReg(CMD_REG, ACC_NORMAL_MODE_CMD);
+    Bmx160WriteReg(CMD_REG, GYRO_NORMAL_MODE_CMD);
+    Bmx160WriteReg(CMD_REG, MAG_NORMAL_MODE_CMD);
+}
+
+void BMX160::ConfigAcc()
+{
+
+    Bmx160WriteReg(ACC_RANGE_REG, ACC_RANGE_8G);
+    Bmx160WriteReg(ACC_CONF_REG, ACC_ODR_800_OSR4);}
```
Pop0097: } on next line pls

Autopilot/Src/imu.cpp
```diff
@@ -0,0 +1,219 @@
...
+        return;
+    }
+
+    int16_t *intImuDataPtr = (int16_t *) &(rawImuData[1]); // first byte is garbage. It's just what was on the line when we asked the IMU for data, which it started sending as of the second byte.
+
+    int16_t magx = *(&(intImuDataPtr[0]));
+    int16_t magy = *(&(intImuDataPtr[1]));
+    int16_t magz = *(&(intImuDataPtr[2]));
+    int16_t gyrx = *(&(intImuDataPtr[4])); // missing 3 is not a mistake, there is a field here named RHall which has something to do with the magnetometer. But I can't quite figure that out yet.
+    int16_t gyry = *(&(intImuDataPtr[5]));
```
Pop0097: Curious to learn more about what is going on here.

You are getting the address of the nth element of the array and then dereferencing it. Why can you not just do `int16_t gyro = intImuDataPtr[4];`?

Autopilot/Inc/imu.hpp
Fberbari: deprecated now. Unit tests and simulation now hook in at the level of sensor fusion, so there's never any need to mock against drivers.

Autopilot/Src/imu.cpp
Fberbari: ya, it's the slave select for the IMU chip. Tried getting the SPI hardware to do it, but it kinda deasserts it at weird times and messes up the transmission. People on the internet recommended to just do it in software like this. 

If it's unclear, I could wrap it inside a litte helper function: Assert_SlaveSelect(). DeAssert_SlaveSelect() ?

Autopilot/Src/imu.cpp
Fberbari: good point. lmao, I think this was an artifact from my previous implementation of this driver.

Autopilot/Src/imu.cpp
Pop0097: nah all good. Comment explains it :))

Pop0097 (approved):
Looks good to me!

PR #126, opened 2021-06-08
Test  compilation failures now force Travis to fail

Fberbari:


Pop0097 (approved):
Tested locally and build halts as expected. Nice!

PR #127, opened 2021-06-10
implemented Interchip 

SnackkOverflowError:
Interchip autopilot is called every 20 ms in a freertos thread (not in this PR, it will be added in another PR). 

Interchip Safety is called once before the main loop, and each interchip callback calls the HAL_SPI_TransmitReceive function, which keeps safety listening and sending data.

Safety has functions to get PWM, and get/setsafety level.
Autopilot has functions to set PWM and get safety level.


interchip-Autopilot/Core/Inc/Interchip.hpp
```diff
@@ -0,0 +1,27 @@
+/*
```
Fberbari: You meant to get rid of all these files, right ?

Safety/boardfiles/Safety_ZeropilotRev1.ioc
```diff
@@ -24,22 +24,22 @@ Mcu.Name=STM32F030RCTx
...
-Mcu.Pin10=PC8
-Mcu.Pin11=PC9
-Mcu.Pin12=PA8
-Mcu.Pin13=PA9
-Mcu.Pin14=PA10
-Mcu.Pin15=PA11
-Mcu.Pin16=PA13
-Mcu.Pin17=PA14
-Mcu.Pin18=PA15
-Mcu.Pin19=PC10
```
Fberbari, line 36: I'm perplexed.... Why are pins being renamed ?

Safety/Src/interchip_S.cpp
```diff
@@ -1,93 +1,57 @@
...
-    if (transmit_status != HAL_OK) {
-        return 1;
-    }
-    else
-    {
-        return 0;
-    }
+// Starts interchip interrupt. Call this once, not in a loop, each interchip callback
+// sets up another interrupt.
+void sendReceiveData() {
```
Fberbari: THis should be done in some sort of init function, The name is misleading.

Safety/Src/interchip_S.cpp
```diff
@@ -1,93 +1,57 @@
...
-
-static Interchip_StoA_Packet* dataTX;
-static Interchip_AtoS_Packet* dataRX;
-static uint16_t errorCount=0;
-//
-
-uint16_t getErrorCount(){
-  return errorCount;
+// Get the pwm signal that has been recieved.
+int16_t getPWM(int8_t index) {
```
Fberbari: Can you actually just return the whole struct 1 shot ? Makes things a bit easier, also easier in terms of the isDataNew stuff, which this function may as well return.

Autopilot/boardfiles/Autopilot_ZeropilotRev1.ioc
```diff
@@ -71,55 +72,59 @@ Mcu.Pin15=PA0/WKUP
...
-Mcu.Pin21=PF14
-Mcu.Pin22=PF15
-Mcu.Pin23=PG0
-Mcu.Pin24=PG1
-Mcu.Pin25=PE7
-Mcu.Pin26=PE13
-Mcu.Pin27=PB10
-Mcu.Pin28=PB11
-Mcu.Pin29=PB12
+Mcu.Pin20=PA5
```
Fberbari, line 77: really unsure of what's going on here.... Kinda making me nervous.

Autopilot/Src/imu.cpp
```diff
@@ -1,4 +1,5 @@
 #include "imu.hpp"
+#include "Interchip_A.h"
```
Fberbari: This is a brutal solution. Just make a file called "Callbacks.cpp" and stick in AutoPilot/src. I'm making a note in the asana that this should be done, but no need to address now, cuz we simply don't have the luxury, but the callbacks file shoiuld just call some appropraite callback function inside each module.

Fberbari (commented):
There's a lot of cosmetic stuff... But not now, after comp. This looks good, but do you know anything about why cubeMX is renaming pins ?? That just seems really weird. Also gotta get rid of all the interchip-autopilot crap

interchip-Autopilot/Core/Inc/Interchip.hpp
SnackkOverflowError: did not realize I commited those, whoops

Autopilot/boardfiles/Autopilot_ZeropilotRev1.ioc
SnackkOverflowError: Me too actually, Ill investigate this further

Autopilot/Src/imu.cpp
SnackkOverflowError: Im not sure if my fix is more or less hacky than what is currently there, but now the callback code is all in Callbacks.cpp

Safety/Src/interchip_S.cpp
SnackkOverflowError: oh yeah I hadnt considered the isDataNew stuff, will do. Should I be returning the entire struct or just the PWM array?

SnackkOverflowError: it now returns the array

Autopilot/Src/Callbacks.cpp
```diff
@@ -0,0 +1,12 @@
+#include "Interchip_A.h"
+#include "imu.hpp"
+#include "spi.h"
+
+void HAL_SPI_TxRxCpltCallback(SPI_HandleTypeDef *hspi)
+{
+    if(hspi == &hspi1) {
```
Fberbari, line 7: This is the right idea. Not hacky, it's good.

For reference, the really proper solution is to have each module pass the callback module a function pointer to call. That way, the callback module doesn't need to #include anyone and also that callback prototypes don't need to be exposed to the higher level. but that's of very low value and this solution is almost just as good.

Autopilot/Inc/Interchip_A.h
```diff
@@ -10,12 +10,14 @@
...
 extern "C" {
 #endif
+ 
 
+void Interchip_Init();
 void Interchip_Run();
-int16_t *Interchip_GetPWM(void);
-void Interchip_SetPWM(int16_t *data);
+void Interchip_SetPWM(int index,int data);
 uint16_t Interchip_GetAutonomousLevel(void);
```
Fberbari, line 21: Each of these will need a comment to explain it. As we do with all our function prototypes.

Autopilot/Inc/imu.hpp
```diff
@@ -84,4 +84,6 @@ class BMX160: public IMU{
 
 };
 
+void ImuTxRxCallback();
```
Fberbari, line 92: Add this above this prototype pls:

        /**
         * Callback for exclusive use by the SPI ISR.
         * Users of this module should never call this function.
         * Any manipulation of this function by the higher level voids all guarantees made by this module.
         * */

Fberbari (commented):
Sweet, couple more small things. the only hurdle is that IOC garbage, but I'm gonna take a look at that as well.

Common/Inc/Interchip.h
```diff
@@ -1,17 +1,12 @@
...
-} Interchip_StoA_Packet;    //Safety to Autopilot packet
 
 typedef struct {
+	uint16_t safetyLevel;
 	int16_t PWM[12];
-	uint16_t autonomous_level;
-} Interchip_AtoS_Packet;    //Autopilot to Safety packet
+} Interchip_Packet;  
 
 #endif
```
Pop0097, line 11: newline . See if there are any other places where the "do not enter" sign appears (we really need to automate this lol)

Safety/Src/interchip_S.cpp
```diff
@@ -1,93 +1,57 @@
...
+	txData.safetyLevel = 13; 
 }
 
-#endif
-
+// spi callback that starts up another interrupt.
+void HAL_SPI_TxRxCpltCallback(SPI_HandleTypeDef* hspi) {
+	// Start Listening after the previous exchange.
+	HAL_SPI_TransmitReceive_IT(&hspi1,(uint8_t *)&txData,(uint8_t *)&rxData, sizeof(Interchip_Packet));
+	dataNew = true;
```
Pop0097, line 56: smart move

Pop0097 (commented):
Not much for me to say beyond what Anthony mentioned.

Fberbari (approved):

Pop0097 (approved):
This is tested to work and the code looks good to me. Let's integrate it and see how it flies!

PR #128, opened 2021-06-11
Adding thread that calls IMU to measure at periodic intervals (Draft)

upadhyaydhruv:


Autopilot/IMU/Src/IMUInterface.cpp
```diff
@@ -0,0 +1,8 @@
+#include "IMUInterface.h"
```
Fberbari: I don't like the idea of creating a folder specifically for the IMU. Just toss these files in inc and src. Maybe some day we can dedicate a folder to drivers and there interfaces.

Fberbari (commented):
This looks fine, but you should address the merge conflicts (probs forgot to rebase)

Autopilot/boardfiles/Src/freertos.c
```diff
@@ -92,26 +93,28 @@ static const int PERIOD_PATHMANAGER_MS = 100;
 static const int PERIOD_TELEMETRY_MS = 100; 
 static const int PERIOD_SENSORFUSION_MS = 200; 
 static const int PERIOD_INTERCHIP_MS = 20;
+static const int PERIOD_IMU_MS = 5; // Needs to be called at 200 Hz. 
 
-static bool catastrophicFailure = false;
+static volatile bool catastrophicFailure = false;
```
Pop0097, line 99: Aadi added this to his pr which was just merged so there will be merge conflicts... should not be too bad to solve tho looking at the affected files

Autopilot/IMU/Src/IMUInterface.cpp
Pop0097: Agreed. Driver files are starting to really pile up

Pop0097 (commented):
Looks good so far. Please fix merge conflicts and move IMU files to our dedicated Inc and Src folders.

Autopilot/IMU/Src/IMUInterface.cpp
Pop0097: In fact, let's add this to this doc: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1780187163/Post-Competition

upadhyaydhruv:
For some reason, when I try to test it on the board the LEDs refuse to toggle. I have a feeling its not due to the changes in the freertos.c because it still doesn't work when I comment out the newly added IMU thread and its references (effectively reverting it to its original state at which I knew it worked). Would any of you have suggestions?

upadhyaydhruv:
This PR wasn't marked as draft and still has a few changes to go through, I'll re-open a new one for this issue when its ready

PR #131, opened 2021-06-11
Implemented Airspeed

SnackkOverflowError:
airspeed driver uses adc 3.

Polls the adc 10 times and averages the results.

Accuracy of ~1 m/s based on my sketchy testing.


Autopilot/Inc/airspeed.hpp
```diff
@@ -13,6 +13,9 @@
 **/
 
 
+#include <cstdint>
```
Pop0097: I'd put this within the compiler guard (After line 20)

Autopilot/Inc/airspeed.hpp
```diff
@@ -55,6 +58,58 @@ class airspeed {
...
+        float airspeed = 0;
+        float utcTime = 0;
+        int sensorStatus = 0;
+        bool dataNew = false;
+        
+};
+
 class dummyairspeed: public airspeed{
     public:
         static dummyairspeed* GetInstance(); //Adding this early so the build works
```
Pop0097: Should be surrounded by some `#if` tags. I'd refer to other sensor drivers to see how this is done

Autopilot/Src/imu.cpp
```diff
@@ -45,7 +45,7 @@
 #define MAG_IF_3_DATA_MODE 0x02
 #define MAG_IF_2_DATA_MODE 0x4C
 #define MAG_IF_1_DATA_MODE 0x42
-#define MAG_REFRESH_50_HZ 0x07  // TODO mag should refresh at 200 Hz as well, but I cant seem to get that to work.
```
Pop0097: Nice catch, but perhaps not best to include with this PR since it is concerning airspeed. You could ask Dhruv U to add it in since his PR concerns itself with IMU.

Pop0097: ngl tho, after competition when all the work is done we should take a few days to just clean up everything. Then everyone can rebase and have a clean env. to start new tasks on

Autopilot/Inc/airspeed.hpp
```diff
@@ -55,6 +58,58 @@ class airspeed {
...
+         * Moves values from the variables to the airspeed data struct.
+         * Sets the dataNew variable to false after that.
+         */
+        void GetResult(airspeedData_t &Data);
+
+        /**
+         *  Triggers interrupt for new airspeed measurement - stores 
+         *  raw data in variables. This function is blocking, as polling for the adc has
+         *  a timeout of 50 ms.
+         */
```
Pop0097: If this is blocking our entire program, 50 ms may be a bit concerning since this will be called often. @Fberbari what are your thoughts?

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,92 @@
+#include "airspeed.hpp"
```
Pop0097, line 1: I really like how you named the file. We may have to adopt this for all sensors.

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "adc.h"
+#include <stddef.h>
+
+#include <cmath>
+
+#define VS 5
+#define DENSITY_AIR 1.2754
+#define ADC_RESOLUTION 4096
+#define KPA_TO_PA 1000
```
Pop0097, line 10: PA_TO_KPA (KPA_TO_PA is 0.001)

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,92 @@
...
+    //constructor
+    HAL_StatusTypeDef status = HAL_ADC_Start(&hadc3);
+    if(status != HAL_OK) {
+        sensorStatus = 1;
+    }
+
+    this->CalcOffset();
+}
+
+MPXV7002DP* MPXV7002DP::s_Instance = NULL;
```
Pop0097: nullptr

Also, can this be put above the GetAverage method please

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+MPXV7002DP* MPXV7002DP::GetInstance() {
+    if(!s_Instance) {
+        s_Instance = new MPXV7002DP;
+    }
+    return s_Instance;
+}
```
Pop0097, line 29: If you are interested in a more C++ style of doing singletons, you can refer to Anthony's imu driver in devel

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (! dataNew)
+    {
+        Data.isDataNew = false;
+        return;
+    }
+    Data.airspeed = this->airspeed;
+    Data.isDataNew = this->dataNew;
+    Data.sensorStatus = this->sensorStatus;
+
+    this->dataNew = false;
```
Pop0097: don't ever need to have this-> 

In C++ when you write a variable idenifier, the compiler searches for the variable in the following order. If I recall correctly, it goes: 
1. local variable (declared within the function)
2. parameter of the function
3. field in defining class (MPXV7002DP)
4. field in inheritance ancestor of defining class (airspeed)
5. global variable

Autopilot/Inc/airspeed.hpp
```diff
@@ -55,6 +58,58 @@ class airspeed {
  * Derived classes
  **********************************************************************************************************************/
 
+class MPXV7002DP : public airspeed {
```
Pop0097, line 61: Note for future, make all classes use upper CammelCase. 

If you could add this to the future tasks doc that would be amazing!

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,92 @@
...
+	uint16_t avg = GetAverage();
+	this->offset = avg - BASELINE;
+
+}
+
+float MPXV7002DP::GetDiffPressure() {
+	float inputVoltage = GetVoltage();
+
+    //formula from datasheet page 6
+	float diffPressure = KPA_TO_PA *( (5 * inputVoltage/VS) - 5.0/2.0);
```
Pop0097: spaces on both sides of `*`

Pop0097 (commented):
Great work so far

Autopilot/Src/imu.cpp
SnackkOverflowError: I don't remember adding this. and have no clue what it does. So I'm going to undo the change.

Autopilot/Inc/airspeed.hpp
SnackkOverflowError: The poll for conversion function to poll the adc should work with a lower timeout, what would be a good value? 10-20 ms? I'll test to find the lowest value and update it.

Autopilot/Src/MPXV7002DP.cpp
SnackkOverflowError: I think I might be confused. 1 KPa = 1000 Pa, so to go from KPa to Pa it should be Kpa * 1000, right?

Autopilot/Src/imu.cpp
Pop0097: Oh, you're just fixing a typo in the comment

Autopilot/Src/MPXV7002DP.cpp
Pop0097: I am stupid. Please ignore this

Autopilot/Inc/airspeed.hpp
SnackkOverflowError: Im not sure if dummyairspeed is used for simulation or unit testing, as there is a mock airspeed for testing and a SimulatedAirspeed for simulation. The only reference to this class has been commented out, so I'll just delete it.

Autopilot/Inc/airspeed.hpp
```diff
@@ -58,7 +58,7 @@ class airspeed {
  * Derived classes
  **********************************************************************************************************************/
 
-class MPXV7002DP : public airspeed {
+class MPXV7002DP : public Airspeed {
```
Pop0097: I would not change the Airspeed class name now because it will mess up a lot of stuff in sensor fusion, etc. Keep it as was. We will change it later.

Pop0097: Screen Shot 2021-06-11 at 8 10 15 PM

Actually looking at this scratch what I said lol. Just need to change this class name

SnackkOverflowError: sounds good, it has already started to break stuff lol

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if(!s_Instance) {
+        s_Instance = new MPXV7002DP;
+    }
+    return s_Instance;
+}
+
+uint16_t MPXV7002DP::GetAverage() {
+	uint16_t avg = 0;
+	for(int i = 0; i < 10; i++) {
+		HAL_ADC_PollForConversion(&hadc3, 10);
```
Pop0097: nice, so 10 ms was the lowest time you could get?

Autopilot/Src/imu.cpp
```diff
@@ -45,7 +45,7 @@
...
+#define MAG_REFRESH_50_HZ 0x07  // TODO mag should refresh at 200 Hz as well, but I cant seem to G that to work.
```
Pop0097: If you could revert this back to how it was and exclude it from the PR that would be great. Simply copying and pasting the file (raw) from WARG/devel will do lol.

Pop0097 (commented):
looking good. Almost ready for approval

Autopilot/Src/MPXV7002DP.cpp
SnackkOverflowError: That was the lowest I went to, Ill try to get it lower.

SnackkOverflowError:
image

-c flag builds

Autopilot/Inc/airspeed.hpp
```diff
@@ -69,11 +70,15 @@ class MPXV7002DP : public airspeed {
         void GetResult(airspeedData_t &Data);
 
         /**
-         *  Triggers interrupt for new airspeed measurement - stores 
-         *  raw data in variables. This function is blocking, as polling for the adc has
-         *  a timeout of 50 ms.
+         *  Don't use this function, the GetResult function is non blocking.
          */
         void Begin_Measuring();
```
Pop0097: Can you please write it as `void Begin_Measuring() {}`. That way it is defined as empty

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -41,19 +39,13 @@ uint16_t MPXV7002DP::GetAverage() {
...
     Data.sensorStatus = sensorStatus;
-
-    dataNew = false;
 }
 
 void MPXV7002DP::Begin_Measuring() {
+    
     airspeed = GetAirspeed();
     dataNew = true;
 }
```
Pop0097: We don't need this anymore, right? If so, remove it and refer to my comment in `airspeed.hpp`

Autopilot/Src/imu.cpp
Pop0097: can you just fix this typo in your code and recommit? If you do this file won't be included in your PR. 

replace the `G` with `get`

Autopilot/Inc/airspeed.hpp
```diff
@@ -13,9 +13,13 @@
 **/
 
 
+
 #ifndef AIRSPEED_HPP
 #define AIRSPEED_HPP
 
+#include <ctime>
```
Pop0097: could you include this in the cpp file if it is not used here?

Autopilot/Src/MPXV7002DP.cpp
```diff
@@ -0,0 +1,83 @@
...
+    Data.airspeed = GetAirspeed();
+    Data.isDataNew = true;
+    Data.sensorStatus = sensorStatus;
+}
+
+void MPXV7002DP::Begin_Measuring() {
+    
+    airspeed = GetAirspeed();
+    dataNew = true;
+}
```
Pop0097: remove since we are not using this method

Pop0097 (commented):
Some quick comments. Then we can approve :))

Pop0097 (approved):
Looks good to me

PR #132, opened 2021-06-15
Created UTC timestamping  functionality

kgrewal26:
A few things I should probably note:

Not entirely too sure if this would work, I mainly based it off my conversation with Dhruv and Tony.

The timer setup is the thing I'm most concerned about - I'm really not sure if I did it right.

Also, the period of the timer hasn't been set yet


Autopilot/Inc/timeStampingPOGI.hpp
```diff
@@ -0,0 +1,25 @@
...
+    private:
+
+        timeStampingPOGI(); //constructor cant be called - singleton
+        static timeStampingPOGI* s_Instance;
+        uint32_t recentUTCTime;
+        void initTimer();
+
+};
+
+#endif
```
Pop0097: If you see this "do not enter" type sign, it means you are missing a neweline character at the end of your file. To fix if you just have to go to the bottom of the file and hit enter once or twice :))

Autopilot/Inc/timeStampingPOGI.hpp
```diff
@@ -0,0 +1,25 @@
+#ifndef TIME_STAMPING_POGI_HPP
+#define TIME_STAMPING_POGI_HPP
+
+#include <cstdint>
+#include <cstdlib>
+#include "gps.hpp"
+
+class timeStampingPOGI {
```
Pop0097: `TimeStampingPOGI`

Pop0097: we use upper CammelCase for classes :))

Autopilot/Inc/timeStampingPOGI.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+    public:
+
+        static timeStampingPOGI* GetInstance();
+        void setGPSTime(GpsData_t* gpsData); //need to change data type depending on what sensorFusion sends me
+        uint32_t getUTCTimeMS();
```
Pop0097: If you could put comments over some of the methods using the following strucuture, that would be great:

```cpp
/**
* Explain what function does
*/
void functionName();
```

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,56 @@
...
+    HAL_TIM_Base_Init(&htim4);
+    HAL_TIM_Base_Start(&htim4);
+
+}
+
+
+//just typical singleton stuff
+
+//Allocates and initializes the static data member pointer (not the object itself)
+timeStampingPOGI* timeStampingPOGI::s_Instance = NULL;
```
Pop0097: `NULL` is C, we use `nullptr` in C++ :))

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+//creates the instance if it doesnt exist & returns the pointer to the instance
+timeStampingPOGI* timeStampingPOGI::GetInstance(){
+
+    if(s_Instance == NULL){
```
Pop0097: `nullptr`

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,56 @@
...
+        recentUTCTime = gpsData->utcTime;
+    }
+
+}
+
+
+//call this to receive the milliseconds since midnight (in UTC)
+uint32_t timeStampingPOGI::getUTCTimeMS(){
+    
+    return (recentUTCTime*1000) + __HAL_TIM_GET_COUNTER(&htim4);
```
Pop0097: space around `*` please

Autopilot/boardfiles/Src/tim.c
```diff
@@ -95,6 +96,42 @@ void MX_TIM11_Init(void)
...
+/* TIM4 Init Function */
+void MX_TIM4_Init(void)
+{
+  
+  TIM_IC_InitTypeDef sConfigIC = {0};
+
+  htim4.Instance = TIM4;
+  htim4.Init.Prescaler = 0;
+  htim4.Init.CounterMode = TIM_COUNTERMODE_UP;
+  htim4.Init.Period = 65535; //STILL NEEDS TO BE SET
```
Pop0097: a 4 bit timer cannot go up to 65535. You need at least a 16 bit timer for this (TIM16)

Autopilot/boardfiles/Src/tim.c
```diff
@@ -95,6 +96,42 @@ void MX_TIM11_Init(void)
...
 }
 
+/* TIM4 Init Function */
+void MX_TIM4_Init(void)
+{
+  
+  TIM_IC_InitTypeDef sConfigIC = {0};
+
+  htim4.Instance = TIM4;
+  htim4.Init.Prescaler = 0;
```
Pop0097: I believe prescalar needs to at least be 1.

Also, what period are you trying to achieve?

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+
+void timeStampingPOGI::setGPSTime(GpsData_t* gpsData){
+
+    //check if its new and set the recentUTCTime to it
+    if (gpsData->timeIsNew){
+        __HAL_TIM_SET_COUNTER(&htim4, 0);
+        recentUTCTime = gpsData->utcTime;
+    }
+
```
Pop0097: The GPS data comes in as `HHMMSS.00` (24 hour clock). You will need to convert it to UTC time

Autopilot/Src/timeStampingPOGI.cpp
Pop0097: Is the counter counting in milliseconds. If so this is a really smart move

Pop0097 (commented):
Good start!

Autopilot/Src/timeStampingPOGI.cpp
kgrewal26: Yup - that was my intention. I can just return milliseconds since midnight or manipulate it into whatever format is needed.

The only thing I haven't set yet is the period for the timer - gonna likely do that tomorrow.

Autopilot/boardfiles/Src/tim.c
kgrewal26: That was one thing I wasnt too sure about yet - I'm gonna think it over tomorrow and recommit

kgrewal26:
Fixed some of the formatting issues. A few things still need to be done:

Changing up the period & switching from tim4 to something else

Will hopefully have that up by tomorrow!

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -1,10 +1,10 @@
 
-#include "timeStampingPOGI.hpp"
+#include "TimeStampingPOGI.hpp"
```
Pop0097: should be `timeStampingPOGI.hpp`. I am surprised this built on your local... what OS are you using?

kgrewal26: hmm, that's weird. I renamed the file to TimeStampingPOGI on my local end, I guess those changes didn't sync through git?

kgrewal26:
Switched to timer12 instead of 4 (this way we can count to at least 1000 with some cushion)

I set the period to 0.001 (1000hz0. The idea behind this is each tick represents 1 millisecond. Let me know if this works because I'm not even entirely sure.

Autopilot/Src/timeStampingPOGI.cpp
Pop0097: Still need to change the file name in this `#include`

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,57 @@
...
+    HAL_TIM_Base_Init(&htim12);
+    HAL_TIM_Base_Start(&htim12);
+
+}
+
+
+//just typical singleton stuff
+
+//Allocates and initializes the static data member pointer (not the object itself)
+TimeStampingPOGI* TimeStampingPOGI::s_Instance = nullptr;
```
Pop0097: declare global variables at the top of the file please

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+
+void TimeStampingPOGI::setGPSTime(GpsData_t* gpsData){
+
+    //check if its new and set the recentUTCTime to it
+    if (gpsData->timeIsNew){
+        __HAL_TIM_SET_COUNTER(&htim12, 0);
+        recentUTCTime = gpsData->utcTime;
+    }
```
Pop0097: You still need to make a method to convert the GPS time to UTC time in seconds since midnight

I want that method to do something like this calculator: https://www.tools4noobs.com/online_tools/hh_mm_ss_to_seconds/

Autopilot/boardfiles/Src/tim.c
```diff
@@ -95,6 +96,44 @@ void MX_TIM11_Init(void)
...
+{
+  
+  TIM_IC_InitTypeDef sConfigIC = {0};
+
+  htim12.Instance = TIM12;
+  htim12.Init.Prescaler = 1;
+  htim12.Init.CounterMode = TIM_COUNTERMODE_UP;
+
+  //this period represents 1000hz - i.e every tick represents a millisecond
+  htim12.Init.Period = 0.001;
```
Pop0097, line 110: if you think back to the bootcamp, the period is the value that your timer will count until.

You want your counter to run at 1000 Hz, so that means your `htim12.Init.Period` should be 1000 if you want each tick to be a milisecond. 

Also, you need to set the prescalar correctly. Our CPU (STM32F765ZG) runs at 216 MHz, so your prescalar must be 216000

Autopilot/boardfiles/Src/tim.c
```diff
@@ -95,6 +96,44 @@ void MX_TIM11_Init(void)
 
 }
 
+/* TIM12 Init Function */
+void MX_TIM12_Init(void)
```
Pop0097, line 100: Ignore this if you have already configured this timer using CubeMX

Right now you have created these timers, but I don't think our chip will be able to run them since we haven't configured them. What you will have to do is use CubeMX (https://www.st.com/en/development-tools/stm32cubemx.html) and add your configurations to our `Autopilot_ZeropilotRev1.ioc` file found in `./boardfiles`. 

After configuration, thse methods will be regenerated and you will need to verify that they are correct. You will also need to include the `.ioc` file in this PR. 

Screen Shot 2021-06-18 at 6 21 44 PM

Pop0097 (commented):
Thanks for getting this in. I have a few comments. 

Do you have a dev board that you can test this on?

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -7,9 +7,9 @@
 TimeStampingPOGI::TimeStampingPOGI(){
 
     //initializing and starting timer
-    MX_TIM4_Init();
-    HAL_TIM_Base_Init(&htim4);
-    HAL_TIM_Base_Start(&htim4);
+    MX_TIM12_Init();
```
Pop0097: You don't need this. If you use CubeMX to configure the ioc this function will be called in `./boardfiles/src/main.c`

Autopilot/Src/timeStampingPOGI.cpp
kgrewal26: oops yeah, not sure how but it completely slipped my mind

Autopilot/boardfiles/Src/tim.c
kgrewal26: Also not sure how this slipped my mind - I must've autopiloted. (no pun intended)

Autopilot/boardfiles/Src/tim.c
kgrewal26: Sounds good - I do remember having trouble with the .ioc file in the past but I'll let you know asap if I run into troubles

Autopilot/boardfiles/Src/tim.c
kgrewal26: > Thanks for getting this in. I have a few comments.
> 
> Do you have a dev board that you can test this on?

Unfortunately I do not. Is there any way some else can test this for me?

kgrewal26:
> Thanks for getting this in. I have a few comments.
> 
> Do you have a dev board that you can test this on?

Unfortunately not, is there a way someone could test it for me?

Autopilot/Src/timeStampingPOGI.cpp
Pop0097: below the #include and #defines tho :))

Pop0097:
Here are the people who currently have dev boards. You can ask around and see if someone is willing:

- Aadi
- Dhruv
- Gordon
- Anni
- Anthony 

Your best bet is probably Aadi if I recall correctly

kgrewal26:
> Here are the people who currently have dev boards. You can ask around and see if someone is willing:
> 
> * Aadi
> * Dhruv
> * Gordon
> * Anni
> * Anthony
> 
> Your best bet is probably Aadi if I recall correctly

 I'm not exactly sure how I'd test something like this.... is it safe to assume they'd know?
 
 Also, the module isn't exactly complete until sensor fusion passes me in the GPS struct, and IIRC that hasn't been done yet.

kgrewal26:
A few new things...

Switched over to timer14 (this hasn't been configured in ioc yet!)

Fixed the way I set the recentUTCTime (accounted for HHMMSS.00 format)

And other small little details.

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -0,0 +1,64 @@
...
+    return s_Instance;
+}
+
+
+void TimeStampingPOGI::setGPSTime(GpsData_t* gpsData){
+
+    //check if its new and set the recentUTCTime to it
+    if (gpsData->timeIsNew){
+
+        __HAL_TIM_SET_COUNTER(&htim12, 0);
```
Pop0097: htim14

If you are changing a variable name, a good practice is to always check with CTRL-F if you have changed all instances ;))

Autopilot/Src/timeStampingPOGI.cpp
```diff
@@ -1,24 +1,20 @@
...
+TimeStampingPOGI* TimeStampingPOGI::s_Instance = nullptr;
 
 TimeStampingPOGI::TimeStampingPOGI(){
 
     //initializing and starting timer
-    MX_TIM12_Init();
-    HAL_TIM_Base_Init(&htim12);
-    HAL_TIM_Base_Start(&htim12);
+    HAL_TIM_Base_Init(&htim14);
+    HAL_TIM_Base_Start(&htim14);
```
Pop0097, line 12: tests failing because htim14 not present. Will be fixed once we set up ioc file :))

PR #138, opened 2021-06-19
GPS fusion and Mahony algorithm

antimonyB:


antimonyB:
./tools/build.bash -c
image

./tools.build.bash -c -t
image
And all tests OK

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -50,22 +51,38 @@ static airspeed *airspeedObj;
...
-  //airspeedObj = TestAirspeed::GetInstance();
+    imuObj = TestIMU::GetInstance();
+    gpsObj = TestGps::GetInstance();
+    //altimeterObj = TestAltimeter::GetInstance();
+    //airspeedObj = TestAirspeed::GetInstance();
 #endif
+
+    //Set initial state to be unknown
+    for(int i = 0; i < NUM_KALMAN_VALUES; i++) iterData.prevX[0] = 0;
+    iterData.prevX[0] = 0;
```
Pop0097: I really dislike these for loops. Can you please use the `{}` brackets?

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -50,22 +51,38 @@ static airspeed *airspeedObj;
...
+    iterData.prevX[2] = 0;
+    iterData.prevX[4] = 0;
```
Pop0097: Didn't you already do this on line 78?

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -75,6 +92,36 @@ void localToGlobalAccel(IMUData_t *imudata, float *u)
...
 }
 
+//Map gps position to xy coords using the reference location as the origin.
+//wikipedia.org/wiki/Geographic_coordinate_system
+double * gpsToCartesian(double lat, double lng){
+    const double RAD_LAT = REF_LAT*ZP_PI/180;
+    static double xy[2] = {0};
+
+    //Latitude
+    xy[0] = (lat - REF_LAT) * (111132.92 - 559.82*cos(2*RAD_LAT) + 1.175*cos(4*RAD_LAT) - 0.0023*cos(6*RAD_LAT));
```
Pop0097: You should use `constexpr` or `#define` for these numerical constants so it is more readable.

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -75,6 +92,36 @@ void localToGlobalAccel(IMUData_t *imudata, float *u)
     u[2] = imudata->accz; //Longitudinal acceleration
 }
 
+//Map gps position to xy coords using the reference location as the origin.
+//wikipedia.org/wiki/Geographic_coordinate_system
+double * gpsToCartesian(double lat, double lng){
```
Pop0097, line 100: Waypoint manager has a function that does pretty much the same thing as this, but it uses the haversine method. We should consider abstracting this so both modules can share the code. 

Should add this in the "things to do" doc

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -75,6 +92,36 @@ void localToGlobalAccel(IMUData_t *imudata, float *u)
...
+
+    //Longitude
+    xy[1] = (lng - REF_LONG) * (111412.84*cos(RAD_LAT) - 93.5*cos(3*RAD_LAT) + 0.118*cos(5*RAD_LAT));
+
+
+    return xy;
+}
+
+double * cartesianToGPS(double x, double y){
+    const double RAD_LAT = REF_LAT*ZP_PI/180;
```
Pop0097: 180.0 to be safe from truncation

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -75,6 +92,36 @@ void localToGlobalAccel(IMUData_t *imudata, float *u)
     u[2] = imudata->accz; //Longitudinal acceleration
 }
 
+//Map gps position to xy coords using the reference location as the origin.
+//wikipedia.org/wiki/Geographic_coordinate_system
+double * gpsToCartesian(double lat, double lng){
+    const double RAD_LAT = REF_LAT*ZP_PI/180;
```
Pop0097: 180.0 to be safe from truncation

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -75,6 +92,36 @@ void localToGlobalAccel(IMUData_t *imudata, float *u)
...
+    xy[0] = (lat - REF_LAT) * (111132.92 - 559.82*cos(2*RAD_LAT) + 1.175*cos(4*RAD_LAT) - 0.0023*cos(6*RAD_LAT));
+
+    //Longitude
+    xy[1] = (lng - REF_LONG) * (111412.84*cos(RAD_LAT) - 93.5*cos(3*RAD_LAT) + 0.118*cos(5*RAD_LAT));
+
+
+    return xy;
+}
+
+double * cartesianToGPS(double x, double y){
```
Pop0097, line 114: how have these methods held up in testing?

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -50,22 +51,38 @@ static airspeed *airspeedObj;
 static SFIterationData_t iterData;
 static SFOutput_t SFOutput;
 
+static double REF_LAT;
+static double REF_LONG;
```
Pop0097: define these variables here as well. Otherwise we can get undefined behaviour:

Set them to 0

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -50,22 +51,38 @@ static airspeed *airspeedObj;
...
+static bool gpsIsSet = false;
+
+const int HIGH_COVAR = 10000;
```
Pop0097, line 62: some comments on what these variables represent would be amazing. Also, after comp if you could write documentation explaining your algorithm that would be incredibly helpful

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -153,6 +200,29 @@ SFError_t SF_GetAttitude(SFAttitudeOutput_t *Output, IMUData_t *imudata, airspee
...
+    double gpsCovar;
+    if(gpsdata->dataIsNew && gpsdata->numSatellites >= 3)
+    {
+        gpsCovar = 2;
+        if(!gpsIsSet)
+        {
+            //Set a reference position to help convert between gps data formats.
+            REF_LAT = gpsdata->latitude;
+            REF_LONG = gpsdata->longitude;
+            gpsIsSet;
```
Pop0097: `gpsIsSet = true;`

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -153,6 +200,29 @@ SFError_t SF_GetAttitude(SFAttitudeOutput_t *Output, IMUData_t *imudata, airspee
...
+        }
+    }
+    else
+    {
+        gpsCovar = HIGH_COVAR+1;
```
Pop0097: space between + signs always

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -228,14 +298,14 @@ SFError_t SF_GetPosition(SFPathOutput_t *Output, AltimeterData_t *altimeterdata,
 
     float q[DIM*DIM]=
```
Pop0097, line 309: comment explaining what this array is please

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -153,6 +200,29 @@ SFError_t SF_GetAttitude(SFAttitudeOutput_t *Output, IMUData_t *imudata, airspee
 
 SFError_t SF_GetPosition(SFPathOutput_t *Output, AltimeterData_t *altimeterdata, GpsData_t *gpsdata, IMUData_t *imudata, SFAttitudeOutput_t *attitudedata,  SFIterationData_t *iterdata)
 {
+    //Set currently unused sensors to zero
+    altimeterdata->altitude = 0;
+    imudata->accx = 0;
+    imudata->accy = 0;
+    imudata->accz = 0;
+
+    double baroCovar = HIGH_COVAR+1;
```
Pop0097: You're not changing this value in this method before line 342. As a result, the if statement on line 342 will always be true. Is this intentional?

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -263,26 +333,34 @@ SFError_t SF_GetPosition(SFPathOutput_t *Output, AltimeterData_t *altimeterdata,
...
     };
-    
-    int16_t baroCovar = 1;
-    int16_t gpsCovar;
-    if(gpsdata->numSatellites >= 3)
-    {
-        gpsCovar = 1 + pow(gpsdata->numSatellites, -0.5);
+    //Force sensors to be completely ignored when the covariance is high
+    if(baroCovar >= HIGH_COVAR){
+        h[0] = 0;
```
Pop0097, line 355: where are you declaring h[]?

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -309,12 +387,17 @@ SFError_t SF_GetPosition(SFPathOutput_t *Output, AltimeterData_t *altimeterdata,
 
     //Defines for updating estimate
 
+    double * xyPos = gpsToCartesian(gpsdata->latitude, gpsdata->longitude);
+
     float z[NUM_MEASUREMENTS*1]=
```
Pop0097: space between `*` and `=`

Autopilot/SensorFusion/Src/SensorFusion.cpp
```diff
@@ -357,9 +440,11 @@ SFError_t SF_GetPosition(SFPathOutput_t *Output, AltimeterData_t *altimeterdata,
...
+    double * latLongOut = cartesianToGPS(x[2], x[4]);
+
     Output->altitude = x[0];
-    Output->latitude = x[2];
-    Output->longitude = x[4];
+    Output->latitude = latLongOut[0];
+    Output->longitude = latLongOut[1];
 
     for (int i = 0; i < DIM*1; i++) iterdata->prevX[i] = newX[i];
     for (int i = 0; i < DIM*DIM; i++) iterdata->prevP[i] = newP[i];
```
Pop0097, line 465: please use `{}` brackets

Autopilot/SensorFusion/Src/SensorFusion.cpp
Pop0097: yeah these one-character names are incredibly confusing. Can you please add comments or improve their names?

Autopilot/AttitudeManager/Src/MahonyAHRS.cpp
```diff
@@ -0,0 +1,232 @@
...
+//---------------------------------------------------------------------------------------------------
+// Header files
+
+#include "MahonyAHRS.h"
+#include <math.h>
+
+//---------------------------------------------------------------------------------------------------
+// Definitions
+
+#define sampleFreq	160.0f			// sample frequency in Hz
```
Fberbari: 200 Hz

Autopilot/AttitudeManager/Src/MahonyAHRS.cpp
```diff
@@ -0,0 +1,232 @@
...
+#define twoKpDef	(2.0f * .5f)	// 2 * proportional gain
+#define twoKiDef	(2.0f * 0.1f)	// 2 * integral gain
+
+//---------------------------------------------------------------------------------------------------
+// Variable definitions
+
+volatile float twoKp = twoKpDef;											// 2 * proportional gain (Kp)
+volatile float twoKi = twoKiDef;											// 2 * integral gain (Ki)
+volatile float q0 = 0.52513f, q1 = 0.051342f, q2 = 0.014727f, q3 = 0.84934f;					// quaternion of sensor frame relative to auxiliary frame
```
Fberbari: I'm curious, where do these initial values come from ?

Autopilot/SensorFusion/Src/SensorFusion.cpp
Fberbari: MathConstants.hpp already has these buil in
image
You should use them instad of doing the conversion manually like this

Fberbari (commented):
Couple of things from me. I'd also like to see answers to Dhruv's comments

Autopilot/SensorFusion/Src/SensorFusion.cpp
antimonyB: They can't be initialized since they have already been declared previously. Would std::fill be better?

Autopilot/SensorFusion/Src/SensorFusion.cpp
antimonyB: Latitude and longitude error for the entire algorithm was within 5m using these, but a simpler approximation gave very similar results.

Autopilot/SensorFusion/Src/SensorFusion.cpp
antimonyB: This temporarily ignores the altimeter since testing was focused on GPS.

Autopilot/SensorFusion/Src/SensorFusion.cpp
antimonyB: Just a couple of lines above, though it's collapsed in Github.

Autopilot/AttitudeManager/Src/MahonyAHRS.cpp
antimonyB: An initial rotation I temporarily changed to match testing data. Setting it back to default now.

Autopilot/SensorFusion/Src/SensorFusion.cpp
Pop0097: Alright. And I assume that simpler approximation is the `LAT_DIST`?

Autopilot/AttitudeManager/Src/MahonyAHRS.cpp
```diff
@@ -29,7 +29,7 @@
 
 volatile float twoKp = twoKpDef;											// 2 * proportional gain (Kp)
 volatile float twoKi = twoKiDef;											// 2 * integral gain (Ki)
-volatile float q0 = 0.52513f, q1 = 0.051342f, q2 = 0.014727f, q3 = 0.84934f;					// quaternion of sensor frame relative to auxiliary frame
+volatile float q0 = 1.0f, q1 = 0.0f, q2 = 0.0f, q3 = 0.0f;					// quaternion of sensor frame relative to auxiliary frame
```
Pop0097, line 32: there a particular reason we are settnig these to 0?

Pop0097 (commented):
Looks pretty good to me. I'm ready to approve it today after a few questions are answered

After comp can you write some documentation explaining the math that went into this? That way two or three years from now younger members can understand what's happening here :))

Pop0097 (approved):

Autopilot/AttitudeManager/Src/MahonyAHRS.cpp
antimonyB: If the imu was attached crooked we could specify that here. This is set to the default of no rotation, so the imu is assumed to be lined up with the forward direction.

PR #141, opened 2021-06-20
Altimeter

Fberbari:
Altimeter driver.

Only caveat is that I couldnt get interrupts working. So this is polling. Tests have shown this polling driver takes 125us to complete an I2C transaction, This happens every 5 ms. It's not great, but it will work.

image


Fberbari:
I'd ask reviewers to ignore changes in the FREERTOS and main.c files. They appear as changes here because they are different from the ones in the FlightTestS2021 branch, which are different than the ones in devel. None of this matters. We will be using the ones that Dhruv produces in the end.

Autopilot/Inc/altimeter.hpp
```diff
@@ -1,69 +1,87 @@
...
+
+/**
+ * This union is of no use to the user.
+ * It is for internal use by the module only.
+ * */
+union AltimeterAltitudeRepresentation
+{
+    uint8_t byte[3];
+    int32_t Q16point4;
 };
```
Pop0097, line 33: flex

Autopilot/Inc/altimeter.hpp
```diff
@@ -1,69 +1,87 @@
...
-class MS5637 : public Altimeter {
+class MPL3115A2 : public Altimeter {
     public:
-        MS5637(const MS5637*) = delete; //Apparently if you try to copy a singleton this will give you errors?
-        static MS5637* GetInstance();
+        /**
+         * This module is built as a singleton. Thus to access an MPL3115A2 object, this function must be called.
+         * Only a single MPL3115A2 object will ever be created and will be shared by all callers of this function.
+         * @return      Altimeter        reference to the singleton object.
+         * */
```
Pop0097, line 64: I like this comment. 

Note to self: use this as the standard comment for all singletons

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,136 @@
...
+#define STANDBY_MODE_8_OS 0x98
+#define ACTIVE_MODE_8_OS 0x99
+
+#define MEM_SIZE_1_B 1
+
+/***********************************************************************************************************************
+ * Static variables
+ **********************************************************************************************************************/
+
+static bool dataIsNew;
```
Pop0097, line 25: `= false`

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+}
+
+void MPL3115A2::ConfigAltimeter(void)
+{
+    HAL_NVIC_DisableIRQ(I2C4_EV_IRQn);
+
+    uint8_t txByte;
+
+    txByte = STANDBY_MODE_8_OS;
```
Pop0097: can this be done in one line?

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,136 @@
...
+    HAL_StatusTypeDef stuff = HAL_I2C_Mem_Write(&hi2c4, LEFT_SHIFTED_ALTIMETER_SLAVE_ADDRESS, ALT_CTRL_REG1, MEM_SIZE_1_B, &txByte, 1, 1000);
```
Pop0097: prefer a variable name like `errorCode`

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    }
+
+    altimeterCalibration /= static_cast<float> (nSamplesForReliableAverage);
+
+    altimeterCalibrationFinal = altimeterCalibration;
+
+}
+
+void MPL3115A2::ConfigAltimeter(void)
```
Pop0097, line 107: A quick comment saying what you are configuring will help future gooslings 

From the `txByte` values I am guessing you are in essence turning off and on the altimeter?

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,136 @@
...
+    static MPL3115A2 singleton;
+    return singleton;
+}
+
+void MPL3115A2::Begin_Measuring(void)
+{
+    HAL_I2C_Mem_Read(&hi2c4, LEFT_SHIFTED_ALTIMETER_SLAVE_ADDRESS, ALT_OUT_P_MSB_REG, MEM_SIZE_1_B, rawAltimeter.byte, 1, 10);
+}
+
+static float dataLog[100];
```
Pop0097: declare and define file-global variables at top pls

Autopilot/Src/altimeter.cpp
```diff
@@ -0,0 +1,136 @@
...
+ * Private methods
+ **********************************************************************************************************************/
+
+MPL3115A2::MPL3115A2()
+{
+    HAL_I2C_MspInit(&hi2c4);
+
+    ConfigAltimeter();
+
+    //Calibrate();
```
Pop0097, line 72: Any reason we are not calibrating?

Autopilot/Src/altimeter.cpp
Fberbari: initialized to false in the constructor.

Autopilot/Src/altimeter.cpp
Fberbari: forgot to get rid of this. It was just for some local tests I ran.

Autopilot/Src/altimeter.cpp
Fberbari: To calibrate, the plane has to be in an "ideal" setting. That means level, not moving, and all this kind of stuff.

So what needs to be done is to put the plane in those ideal conditions, calibrate, and store the calibration values in EEPROM for future use. My plan is to do this once on the WATERLOO airstrip and hardcode the values in the module. It doesent make sense to do this every time the micro is fired up because those ideal conditions aren't always guarantedd.

Fberbari: These things aren't as important for the altimeter as they are for something like an IMU, but I'll just do it all one shot.

Autopilot/Src/altimeter.cpp
Fberbari: Good point.

Autopilot/Src/altimeter.cpp
Fberbari: yes it can, but for consistency with the next byte, I figured it was prettier like this

Autopilot/Src/altimeter.cpp
Fberbari: forgot to throw this away as well. My b.

Autopilot/Src/altimeter.cpp
Pop0097: sounds good

Pop0097 (approved):
Tested to work on hardware and it looks good to me too

PR #142, opened 2021-06-21
Simulation Testing

NixonCTChan:
- tested and tuned PID's for path and attitude manager
- testing cruising module
- tested landing and takeoff states

Note: build doesn't pass since there are some telemetry comms issues, which are being fixed in another PR


Autopilot/PathManager/Inc/pathDatatypes.hpp
```diff
@@ -29,6 +30,8 @@ struct _PassbyControl{
 
     double throttlePercent;
     bool throttlePassby;
+    _PassbyControl() : rollPercent(0), rollPassby(false), rudderPercent(0), rudderPassby(false), pitchPercent(0), pitchPassby(false), throttlePercent(0), throttlePassby(false)
```
Pop0097: My rule of thumb:

structs should only be used for plain old data, that is, data only and no methods. If you are adding a constructor I would make this a class.

Autopilot/PathManager/Src/CommsWithTelemetry.cpp
```diff
@@ -56,6 +55,5 @@ bool GetTelemetryCommands(Telemetry_PIGO_t *commands)
         //Indicate that no new commands are available.
         return false;
     }
-    */
     return true;
```
Pop0097, line 63: Are the changes in this file needed? If not let's exclude this file

Autopilot/PathManager/Src/landingTakeoffManager.cpp
```diff
@@ -120,6 +120,15 @@ _LandingPath LandingTakeoffManager::createSlopeWaypoints(const Telemetry_PIGO_t
 _LandingTakeoffOutput LandingTakeoffManager::translateWaypointCommands(const _WaypointManager_Data_Out & outputData)
 {
     _LandingTakeoffOutput controlOutput;
```
Pop0097: You can use value initialization to prevent writing all of these lines of code. You do this by adding `{}`:

```cpp
_LandingTakeoffOutput controlOutput {};
```

Value initialization sets all members to their default value (0 for ints and false for bools)

Autopilot/Simulation/Intercepts/CommsWithTelemetry_intercept_user.cpp
```diff
@@ -21,11 +23,51 @@ void SendPathData(POGI *data)
...
+
+    if(index==800)
+    {
+        userPigo[index].beginLanding = true;
+        userPigo[index].stoppingAltitude = 0;
+        userPigo[index].stoppingDirectionHeading = 90;
+        userPigo[index].stoppingLongitude = -157.91;
+        userPigo[index].stoppingLatitude = 21.3200;
+    }
+
```
Pop0097: I will dm you the tests I used to test the waypoint manager

Autopilot/Simulation/SimDriver/SimMain.cpp
```diff
@@ -1,14 +1,14 @@
...
 
 #include <iostream>
 
 /***********************************************************************************************************************
  * Code
  **********************************************************************************************************************/
 
-#define NUM_SIMULATION_STEPS 1000
-
+#define NUM_SIMULATION_STEPS 2000
```
Pop0097, line 11: make sure this is <= number of elements in PIGO array :))

Autopilot/PathManager/Src/CommsWithTelemetry.cpp
NixonCTChan: its removing the comments put by previous Chief

Autopilot/PathManager/Src/landingTakeoffManager.cpp
NixonCTChan: okok thanks!

Autopilot/PathManager/Inc/cruisingState.hpp
```diff
@@ -10,6 +10,8 @@
 #ifndef CRUISING_STATE
 #define CRUISING_STATE
 
+constexpr int CRUISING_AIRSPEED {30};
```
Fberbari: it was mentioned multiple times that this should be something like 15 m/s, no ? I'm not even sure Spike can achieve 30

Autopilot/PathManager/Src/AutoSteer.cpp
```diff
@@ -12,14 +12,15 @@
...
+static PIDController bankPid{4.2,0,0,0, -MAX_BANK_ANGLE, MAX_BANK_ANGLE}; //5,0,8,0,
 
-static PIDController pitchPid{1, 0, 0, 0, -MAX_PITCH_ANGLE, MAX_PITCH_ANGLE}; // PID gains need to be tuned
+static PIDController rudderPid{0.2,0.2,7,1, -100, 100}; //0.2,0.2,7,1
 
-static PIDController altitudePid{1, 0, 0, 0, -100, 100}; //PID gains need to be tuned
+static PIDController pitchPid{1,0,0,0, -MAX_PITCH_ANGLE, MAX_PITCH_ANGLE}; //1,0,0,0,
 
-static const float RUDDER_SCALING_FACTOR = 0.8f; // should be experimentally determined
+static PIDController altitudePid{110,0,0,0, 0, 100}; //110,0,0,0,
```
Fberbari: this seems massive. A single meter off would drive the throttle to 0 or 100%. That's an incredible strain on the motor.

Fberbari (commented):
Logic verified by sim, which is great. I'm just worried about these incredibly massive PID gains and cruising speeds. I guess they'l have to be messed with in the field anyway. But values that seem reasonable would be a better place to start imo

Autopilot/PathManager/Src/AutoSteer.cpp
NixonCTChan: yup will change. Name should be changed too, since we restructured it to be throttle controlling airspeed.

Autopilot/PathManager/Inc/cruisingState.hpp
NixonCTChan: I'll test some other speeds out

Autopilot/PathManager/Inc/landingTakeoffManager.hpp
```diff
@@ -54,7 +54,7 @@ struct _LandingTakeoffOutput //this structure is used to hold the outputs of eac
     bool useHeading; //this variable is used to determine if heading needs to be controlled, rather than track
     double desiredAltitude;
     double desiredAirspeed;
-    _PassbyControl controlDetails; //this variable is used to store the controls if passby control is needed
+    _PassbyControl controlDetails {}; //this variable is used to store the controls if passby control is needed
```
Pop0097: Don't really initialize in struct declarations

Pop0097 (approved):
Looks alright. We know that this code works from our simulation tests. Some stuff will need to be cleaned up after comp tho.

PR #143, opened 2021-06-22
timeStampingPOGI V2

kgrewal26:
Had to copy and paste everything onto a new branch...

IOC still needs to be configured and it needs to be tested.

Can't be built until IOC is configured.


Autopilot/Inc/timeStampingPOGI.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+class TimeStampingPOGI {
+
+    public:
+
+        /*
+        *Called to either create or get the existing instance of the class
+        *Constructor is private so it only gets called through here (singleton)
+        */
+        static TimeStampingPOGI* GetInstance();
```
Pop0097, line 19: Need a line deleting the constructor so people don't accidentally initialize it. You can look at line 67 of this file: https://github.com/UWARG/ZeroPilot-SW/blob/devel/Autopilot/Inc/gps.hpp

Pop0097 (approved):
Looks good. It was verified on hardware too!

PR #144, opened 2021-06-22
implemented RSSI code

SnackkOverflowError:


Safety/Src/RSSI.cpp
```diff
@@ -0,0 +1,25 @@
+#include "RSSI.hpp"
+#include "gpio.h"
+
+static volatile bool failed = false;
+static volatile int16_t counter = 0;
+
+bool CommsFailed() {
```
Pop0097, line 7: In the future methods like this should start with "get"

Safety/Src/RSSI.cpp
```diff
@@ -0,0 +1,25 @@
+#include "RSSI.hpp"
+#include "gpio.h"
+
+static volatile bool failed = false;
+static volatile int16_t counter = 0;
```
Pop0097, line 5: uint16

Fberbari (approved):
I'm fine with this

Safety/Inc/RSSI.hpp
```diff
@@ -0,0 +1,7 @@
+#ifndef RSSI_HPP
+#define RSSI_HPP
+
+bool CommsFailed();
```
Pop0097, line 4: Comments over functions to explain purpose

Pop0097 (approved):
Looks good. Comments I put should be implemented after competition.

PR #146, opened 2021-06-23
Altimeter and airspeed integration into threads

upadhyaydhruv:
Based on the drivers written by Aadi and Chief, I've added interface implementations and integrated them into the threads written previously. Altimeter is fully complete and tested (threads still work), airspeed ran into a bit of an issue with the initialization of the singleton - "Non-const lvalue reference to type 'airspeed' cannot bind to a temporary of type 'MPXV7002DP ". I asked Aadi but it seems he's still asleep. It's really late and I can't think anymore but hopefully this PR is good enough apart from the airspeed which I'll implement asap tomorrow morning.

I've also copy-pasted Anthony's implementation of the altimeter driver into this branch so my code could build based off of it. The altimeter interface implementation seems a bit iffy since I was confused as to to use the Begin_Measuring(), etc. functions and did it really similarly to the IMU.


Pop0097 (commented):
Looks good so far.

upadhyaydhruv:
> Looks good so far.

Do you think the way the altimeter interface is implemented is legitimate (should I be calling another function instead)?

Also, would you know how to instantiate an airspeed object without the error: "Non-const lvalue reference to type 'airspeed' cannot bind to a temporary of type 'MPXV7002DP"? EDIT: It works as a pointer and not a reference, so we're good

PR #151, opened 2021-08-11
Path Manager Restructure

Pop0097:
Here is the new architecture for path manager. 

Will need to verify with simulation before merging. 

Here's the architecture: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1866989569/Proposed+Redesign


Pop0097:
@sahil-kale

Autopilot/PathManager/Inc/cruisingMode/cruisingModeStages.hpp
```diff
@@ -0,0 +1,34 @@
...
+        void exit(CruisingMode* cruise_mode) { (void) cruise_mode; }
+        static CruisingModeStageManager& getInstance();
+
+    private: 
+        CruisingFlight() : in_hold {false}, going_home {false} {}
+        CruisingFlight(const CruisingFlight& other);
+        CruisingFlight& operator=(const CruisingFlight& other);
+
+        WaypointManager cruising_state_manager;
+        int waypoint_id_array[PATH_BUFFER_SIZE]; // Stores ids of the waypoints in the flight path in the order that they are executed
```
sahil-kale: stdint please

Autopilot/PathManager/Inc/cruisingMode/cruisingModeStages.hpp
```diff
@@ -0,0 +1,34 @@
...
+        static _WaypointManager_Data_In _input_data;
+        static _WaypointManager_Data_Out _output_data;
+        static _CruisingState_Telemetry_Return _return_to_ground;
+        bool in_hold;
+        bool going_home;
```
sahil-kale, line 32: These bools feel awfully specific.... any chance they can be made abstracted?

Autopilot/PathManager/Src/pathModeSelector.cpp
```diff
@@ -0,0 +1,57 @@
...
+    passby_control_output.rollPassby = passby.rollPassby;
+
+    passby_control_output.rudderPercent = passby.rudderPercent;
+    passby_control_output.rudderPassby = passby.rudderPassby;
+
+    passby_control_output.pitchPercent = passby.pitchPercent;
+    passby_control_output.pitchPassby = passby.pitchPassby;
+
+    passby_control_output.throttlePercent = passby.throttlePercent;
+    passby_control_output.throttlePassby = passby.throttlePassby;
```
sahil-kale: If these items are the exact same, why not use memcpy then?

Autopilot/PathManager/Src/takeoffMode/takeoffModeStages.cpp
```diff
@@ -0,0 +1,181 @@
...
+    _output.controlDetails.throttlePercent = 100;
+
+    //making sure landing points are only made once
+    if (!takeoff_mode->getMadeTakeoffPoints()) {
+        //creating waypoints
+        _takeoff_point = LandingTakeoffManager::createTakeoffWaypoint(_input.sensorOutput->latitude, _input.sensorOutput->longitude, _input.sensorOutput->altitude, _input.telemetryData->takeoffDirectionHeading);
+        _path_array[0] = _takeoff_path.initialize_waypoint(_takeoff_point.longitude, _takeoff_point.latitude, _takeoff_point.altitude, PATH_FOLLOW);
+        
+        //10 meters is added to the altitude of the currentLocation waypoint so that is is not ground level
+        _current_location = _takeoff_path.initialize_waypoint(_input.sensorOutput->longitude, _input.sensorOutput->latitude, (_input.sensorOutput->altitude + 10), HOLD_WAYPOINT);
```
sahil-kale, line 55: why specifically 10 meters? Can this be made adjustable

Autopilot/PathManager/Src/landingMode/landingModeStages.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+    //making sure both headings are positive
+    if(difference_in_heading_1 < 0) {difference_in_heading_1 += 360;}
+    if(difference_in_heading_2 < 0) {difference_in_heading_2 += 360;}
+
+    //if the smaller heading is less than 30 degrees, set stage to slope
+    if((fabs(difference_in_heading_1) <= 80) || (fabs(difference_in_heading_2) <= 80)) {
+        land_mode->setCurrentStage(&LandingSlopeStage::getInstance());
+        land_mode->setCurrentStageEnum(LANDING_SLOPE);
+    }
```
sahil-kale, line 102: Coukd you explain this part of the code a little more please?

sahil-kale (changes requested):
Few changes. If there's a doc with the overall architecture of this restructure it'd be good to link

sahil-kale:
Can this be made draft before simulation changes are done?

Autopilot/PathManager/Src/takeoffMode/takeoffModeStages.cpp
Pop0097: @NixonCTChan

Autopilot/PathManager/Src/landingMode/landingModeStages.cpp
Pop0097: @NixonCTChan

Autopilot/PathManager/Inc/cruisingMode/cruisingModeStages.hpp
Pop0097: I can make them 16 bit. Don't want to risk the case where we overflow an 8 bit integer

Autopilot/PathManager/Inc/cruisingMode/cruisingModeStages.hpp
Pop0097: Can you elaborate on what you mean here please? 

The bools are used in the cruisingFlight.cpp file, which calls different functions in waypoint manager depending on these values.

Autopilot/PathManager/Src/pathModeSelector.cpp
Pop0097: oh god, I forgot that existed...

Pop0097: addressed

Autopilot/PathManager/Inc/cruisingMode/cruisingModeStages.hpp
Pop0097: addressed

Autopilot/PathManager/Src/landingMode/landingModeStages.cpp
NixonCTChan: It's almost the same landing code that has been used before for the landing stages, there is documentation on both landing and takeoff if you want to read more about it.

For this part, the heading of the plane is compared with the landing heading. When the plane is in transition and it gets close enough within the heading, we need the plane to transition to slope mode, where more adjustments will be made to get to the desired flight path.

@Pop0097 since we have a magic numbers PR, should probably change the 80 to a const.

Link: https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1607630921/Landing+and+Takeoff

Autopilot/PathManager/Src/landingMode/landingModeStages.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+    //calculating the difference in heading to detect if finished turning (2 differences possible)
+    double difference_in_heading_1 = _input.telemetryData->stoppingDirectionHeading - _input.sensorOutput->track;
+    double difference_in_heading_2 = _input.sensorOutput->track - _input.telemetryData->stoppingDirectionHeading;
+
+    //making sure both headings are positive
+    if(difference_in_heading_1 < 0) {difference_in_heading_1 += 360;}
+    if(difference_in_heading_2 < 0) {difference_in_heading_2 += 360;}
+
+    //if the smaller heading is less than 30 degrees, set stage to slope
```
NixonCTChan, line 98: this should be 80 degrees, my bad, forgot to change it in my last PR

NixonCTChan (commented):
I'll explain isPackage here generally. 

Since the addition of a carry load significantly changes the CG of the aircraft as well as its overall weight, the stall speed would be significantly impacted. To adjust for this, isPackage is used to determine if the CG and weight of the aircraft were changed by the package, which will then determine what speeds we would need to fly for various stages. Since we don't actually have in-person testing with the actual plane, these speeds are somewhat arbitrary for now, but in the future when we can meet in person again, these numbers will need to be redefined to the characteristics of the aircraft.

For example, the landingTakeoff function "slowFlightSpeed" takes in isPackage, and returns a speed that the plane should fly for slow flight, depending on the configuration of the aircraft.


Btw, trying to run simulation on this branch rn. Right before the competition, prior Chief told me to adjust the constants to more reasonable flying IRL, but those values don't really work for simulation. Those would have to be changed back for the sim to run.

Autopilot/PathManager/Src/takeoffMode/takeoffModeStages.cpp
NixonCTChan: yeah we should, @Pop0097 could you make a const and set it to 10 meters? Could call it kTakeoffAltitudeOffset.

PR #155, opened 2021-09-25
Dhruvu/hcsro4 driver

upadhyaydhruv:
Hey guys, this is the driver rewrite for the ultrasonic sensor we plan to use in our drone, the HC-SR04, which is the one found in most generic Arduino kits.

I've written the header file for it and would love some feedback before I implement the source file!

Here's the datasheet for it if you would like to review: https://cdn.sparkfun.com/datasheets/Sensors/Proximity/HCSR04.pdf


Autopilot/Inc/imu.hpp
```diff
@@ -86,6 +86,7 @@ class BMX160: public IMU{
 
 /**
 * Callback for exclusive use by the SPI ISR.
+
```
sahil-kale, line 89: Why?

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include "stm32f765xx.h"
+#include "stm32f7xx_hal_tim.h"
+
+typedef struct ultrasonicData_t {
+    float distance;
+    bool isDataNew; // Records whether the data has been refreshed since the last time getResult was called
+    GPIO_TypeDef * triggerPin;
+    GPIO_TypeDef * echoPin;
+    TIM_HandleTypeDef htim;
+} ultrasonicData_t;
```
sahil-kale: Myabe add the timestamp as well into this

Autopilot/Src/ultrasonic.cpp
```diff
@@ -0,0 +1,63 @@
+#include "ultrasonic.hpp"
+#include "stm32f765xx.h"
+#include "stm32f7xx_hal_tim.h"
+#include "stm32f7xx_hal_gpio.h"
+#include "tim.h"
+#include <cstdint>
+
+#define SPEED_OF_SOUND 0.034f  
+#define TRIG_PIN   5 // temporary value, to be changed when pin configs completed
```
sahil-kale, line 12: where would trig pin be dfined then if it's in the ultrasonic struct?

Autopilot/Src/ultrasonic.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+uint32_t IC_Val1 = 0;
+uint32_t IC_Val2 = 0;
+uint32_t difference = 0;
+uint8_t isFirstCaptured = 0; // Is the first value captured
+uint8_t distance = 0; // computed distance based on the length of the ECHO
```
sahil-kale, line 26: Static please

sahil-kale: Actually, these look like variables that should be part of the class

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,43 @@
...
+        virtual void getDistance(ultrasonicData_t * data) = 0;
+        static ultrasonicData_t data;
+    
+};
+
+class HCSR04 : public ultrasonic {
+    public:
+        HCSR04(const HCSR04*) = delete;
+        void getDistance(ultrasonicData_t * data);
+
```
sahil-kale: I think you're missing a getInstance function here...

Autopilot/Src/ultrasonic.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void HCSR04::trigger(GPIO_TypeDef * triggerPin, uint32_t pinName) {
+    triggerPin -> BSRR |= (1 << TRIG_PIN); // set GPIO pin high
+    this->delayMicroseconds(10); // Delay/wait for 10 us as defined in datasheet
+    triggerPin -> BSRR |= (1 << (16 + TRIG_PIN)); // reset the pin, hence completing the sending of the pulse
+}
+
+void HCSR04::delayMicroseconds(uint32_t us) {
+    int16_t currentTim = htim11.Instance -> CNT;
```
sahil-kale, line 55: why int16_t?

Autopilot/Src/ultrasonic.cpp
```diff
@@ -0,0 +1,63 @@
...
+    while ((htim11.Instance -> CNT - currentTim) < us); // wait the required microseconds before ending
+}
+
+void HCSR04 :: getDistance(ultrasonicData_t * data) {
+    trigger(data -> triggerPin, TRIG_PIN); // Sends trigger pulse 
+    __HAL_TIM_ENABLE_IT(data -> htim, TIM_IT_CC1);
+}
```
sahil-kale: This function feels misleading, it doesn't actually update the data, just triggers the measurement. I'd expect getDistance to update the data, please clarify in comment or rename function

Autopilot/Src/ultrasonic.cpp
```diff
@@ -0,0 +1,63 @@
...
+        else if (isFirstCaptured == 1) {
+            IC_Val2 = HAL_TIM_ReadCapturedValue(htim, HAL_TIM_ACTIVE_CHANNEL_1);
+            isFirstCaptured = 0;
+            __HAL_TIM_SET_COUNTER(htim, 0);
+
+            if (IC_Val2 > IC_Val1) {
+                difference = IC_Val2 - IC_Val1;
+            }
+            else if (IC_Val1 > IC_Val2) {
+                difference = (0xffff - IC_Val1) + IC_Val2; // find difference between val1 and max, and add overflow val2
```
sahil-kale, line 80: Please make the 0xffff a constant of sorts in a define to something.

Also, what are the timing constraints on the timer? Max pulse width is 38ms, so there sohuldn't be an overflow if we're smart about how we start the timer

sahil-kale (changes requested):
I feel as though the singleton design pattern is not correctly implemented here, as the function names seem misleading and "data" is being used to store initial configurations.

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,43 @@
...
+*/
+
+
+#ifndef ULTRASONIC_HPP
+#define ULTRASONIC_HPP
+
+#include "stdint.h"
+#include "stdbool.h"
+#include "stm32f765xx.h"
+#include "stm32f7xx_hal_tim.h"
```
Pop0097, line 16: Let's include this in the .cpp file so any file that includes `ultrasonic.hpp` is still unit testable

Autopilot/Inc/ultrasonic.hpp
Pop0097: yeah, we want to use the singleton pattern here. Take a look at the other sensor drivers for reference

Autopilot/Inc/ultrasonic.hpp
```diff
@@ -0,0 +1,43 @@
...
+    public:
+        HCSR04(const HCSR04*) = delete;
+        void getDistance(ultrasonicData_t * data);
+
+    private:
+        HCSR04();
+        void trigger(GPIO_TypeDef * triggerPin, uint32_t pinName);
+        void delayMicroseconds(uint32_t us);
+
+};
```
Pop0097: ree. newline error

sahil-kale: @Pop0097 Can we get clang-format working for our repo?

PR #157, opened 2021-11-17
Path manager restructure

Gongsta:
This is a pull requests that implements an instructionQueue.

Currently, we can only modify the flight path when we are in cruising mode. The problem with that is that if we are in takeoff mode, and we receive instructions to modify the flight path, these instructions will simply be lost. 

This problem has been fixed by adding an instruction queue (implemented as a linked list) that stores these instructions when we are not in cruising mode. Once we transition to cruising mode, we will check the queue first, and if the queue is not empty, execute those instructions first.


Autopilot/PathManager/Inc/pathModeSelector.hpp
```diff
@@ -34,6 +39,21 @@ class PathModeSelector {
          * @return none
          */
         void execute(Telemetry_PIGO_t telemetry_in, SFOutput_t sensor_fusion_in, IMU_Data_t imu_data_in);
+        
+        /**
+         * Checks whether the instruction queue is empty.
+         * 
+         * @return true if the queue is empty. Else return false
+         */
+        bool instructionQueueIsEmpty();
```
Pop0097: To communicate that this function is specific to cruising mode, maybe name the function something like:

`cruisingEditPathInstrQueueIsEmpty()` or something else if you have ideas

that way if we ever need to add new queues we can easily tell which queue we are calling.

Autopilot/PathManager/Inc/pathModeSelector.hpp
```diff
@@ -34,6 +39,21 @@ class PathModeSelector {
...
+
+        /**
+         * Dequeues the first(a.k.a. most urgent) instruction from the instruction queue. Make sure that queue is not empty
+         * before dequeing by running instructionQueueIsEmpty() beforehand.
+         * 
+         * @return telemetry data
+         */
+        Telemetry_PIGO_t dequeueInstruction();
```
Pop0097: `cruisingEditPathInstrPop()`?

Nice job with the comments :))

Autopilot/PathManager/Inc/pathModeSelector.hpp
```diff
@@ -75,7 +95,7 @@ class PathModeSelector {
          * 
          * @return none
          */
-        void setAltitdeAirspeedInput(AltitudeAirspeedInput_t alt_airspeed_input);
+        void setAltitudeAirspeedInput(AltitudeAirspeedInput_t alt_airspeed_input);
```
Pop0097, line 106: nice typo catch. 

In the future though, it would be better if you actually made a ticket for us to edit it instead of fixing it in this branch. Reason why is by fixing this typo you are changing files that are outside the scope of your task, resulting in potential merge conflicts with other PRs in the future. It's not an issue now since you are the only one touching path manager, but something to note for the future when you work on things for WARG and coop

Autopilot/PathManager/Inc/pathModeSelector.hpp
```diff
@@ -165,6 +185,12 @@ class PathModeSelector {
     private: 
         PathModeSelector();
         static PathModeSelector* singleton;
+        
+        /*
+         * Function to enqueue a new instruction onto the instruction queue.
+         */
+        void enqueueInstruction(Telemetry_PIGO_t newInstruction);
+        instructionQueueNode* first_instr;
```
Pop0097: In your comment for the function here I'd write an @param field to explain what the parameter is.

Also, add a comment to the instructionQueueNode pointer so it's more clear what it is being used for. Maybe renaming it to `cruising_edit_path_first_instr` may be more clear (especially if we add more queues in the future)

Autopilot/PathManager/Src/pathModeSelector.cpp
```diff
@@ -15,21 +15,69 @@ PathModeSelector* PathModeSelector::getInstance() {
...
+    first_instr = nullptr;
 }
 
 #endif
 
 void PathModeSelector::execute(Telemetry_PIGO_t telemetry_in, SFOutput_t sensor_fusion_in, IMU_Data_t imu_data_in) {
+    
+    //Check whether the telemetry_in has an editflightPath command. If so, we enqueue it into our InstructionQueue
+    if (telemetry_in.waypointModifyFlightPathCommand != NO_FLIGHT_PATH_EDIT) {        
+        PathModeSelector::enqueueInstruction(telemetry_in);
```
Pop0097: no need to write `PathModeSelector::`. Since you are actually in a method that is a part of the `PathModeSelector` class, you can access all public and private methods within it.

Also, we really only use `ClassName::` when we are calling static methods from functions outside of the ClassName class. You can read more about it here: https://stackoverflow.com/questions/4365982/how-do-i-call-a-static-method-of-another-class

Autopilot/PathManager/Src/cruisingMode/cruisingModeStages.cpp
```diff
@@ -17,16 +17,25 @@ _CruisingState_Telemetry_Return CruisingFlight::_return_to_ground;
...
-    setReturnValues(&_return_to_ground, cruising_state_manager, edit_error, path_error); // Set error codes
+
+    Telemetry_PIGO_t instr;
+    //Go over all instructions on the instruction queue
+    while (!cruise_mode->getModeSelector()->instructionQueueIsEmpty()) {
+        instr = cruise_mode->getModeSelector()->dequeueInstruction();
+
+        _ModifyFlightPathErrorCode edit_error = editFlightPath(&instr, cruising_state_manager, waypoint_id_array); // Edit flight path if applicable
+        _GetNextDirectionsErrorCode path_error = pathFollow(&instr, cruising_state_manager, _input_data, &_output_data, going_home, in_hold); // Get next direction or modify flight behaviour pattern
+        setReturnValues(&_return_to_ground, cruising_state_manager, edit_error, path_error); // Set error codes
```
Pop0097: These don't need to be in the while loop. `pathFollow()` and `setReturnValues()` only need to be run once when `cruisingFlight::execute()` is called.

Autopilot/PathManager/Src/cruisingMode/cruisingModeStages.cpp
```diff
@@ -17,16 +17,25 @@ _CruisingState_Telemetry_Return CruisingFlight::_return_to_ground;
...
-    _ModifyFlightPathErrorCode edit_error = editFlightPath(&telem_data, cruising_state_manager, waypoint_id_array); // Edit flight path if applicable
-    _GetNextDirectionsErrorCode path_error = pathFollow(&telem_data, cruising_state_manager, _input_data, &_output_data, going_home, in_hold); // Get next direction or modify flight behaviour pattern
-    setReturnValues(&_return_to_ground, cruising_state_manager, edit_error, path_error); // Set error codes
+
+    Telemetry_PIGO_t instr;
+    //Go over all instructions on the instruction queue
+    while (!cruise_mode->getModeSelector()->instructionQueueIsEmpty()) {
+        instr = cruise_mode->getModeSelector()->dequeueInstruction();
+
+        _ModifyFlightPathErrorCode edit_error = editFlightPath(&instr, cruising_state_manager, waypoint_id_array); // Edit flight path if applicable
```
Pop0097: Something to think about: We return an error code when we edit the flight path. What should we do if we get an error while updating the flight path?

Autopilot/PathManager/Src/pathModeSelector.cpp
```diff
@@ -15,21 +15,69 @@ PathModeSelector* PathModeSelector::getInstance() {
...
+        return true;
+    } else {
+        return false;
+    }
+}
+
+Telemetry_PIGO_t PathModeSelector::dequeueInstruction() {
+    ///Assumes that Queue is non empty.
+        instructionQueueNode* newFirstInstruction = first_instr->nextInstruction; //Store temporary new head
+        Telemetry_PIGO_t instruction = first_instr->instruction;
```
Pop0097: These steps are correct

Autopilot/PathManager/Src/pathModeSelector.cpp
```diff
@@ -15,21 +15,69 @@ PathModeSelector* PathModeSelector::getInstance() {
...
+        free(first_instr); //Not sure if this the right way to free variables
```
Pop0097: in C++ to allocate and free memory we use the `new` and `delete` keywords: https://www.cplusplus.com/doc/tutorial/dynamic/

Pop0097: Also, before you delete the memory at first_instr, you have to first set its `first_instr->nextInstruction` to `nullptr` so you don't accidentally delete the next node as well :))

Autopilot/PathManager/Src/pathModeSelector.cpp
```diff
@@ -15,21 +15,69 @@ PathModeSelector* PathModeSelector::getInstance() {
...
+        first_instr = newFirstInstruction; //Update the head of the instructionQueue
```
Pop0097, line 53: good

Autopilot/PathManager/Src/pathModeSelector.cpp
```diff
@@ -15,21 +15,69 @@ PathModeSelector* PathModeSelector::getInstance() {
...
+            latestInstruction = latestInstruction->nextInstruction;
+        };
+
+        //Create new instructionQueueNode
+        instructionQueueNode* newInstructionQueueNode = new instructionQueueNode{};
+        newInstructionQueueNode->instruction = newInstruction;
+        newInstructionQueueNode->nextInstruction = nullptr;
+        
+        //Add new instructionQueueNode to the entire queue
+        latestInstruction->nextInstruction = newInstructionQueueNode;
```
Pop0097, line 92: Looks good. We should write two unit tests for this when the time comes tho :)

Question: What do you think the tests should cover?

Pop0097 (changes requested):
Left some comments. Nice work so far!

Also, I'd recommend looking over our [Style Guide](https://uwarg-docs.atlassian.net/wiki/spaces/ZP/pages/1827635203/Style+Guide) and fixing any inconsistencies you have with it :))

Autopilot/PathManager/Inc/pathModeSelector.hpp
Gongsta: ok sounds good!

Autopilot/PathManager/Src/cruisingMode/cruisingModeStages.cpp
Gongsta: Not sure about this, can we want to break out of the while loop (that dequeues the instructions) so we can run the setReturnValues function ASAP? We don't lose the flight path edit instructions that are stored on the queue, it's just going to be run later since handling the error code is a higher priority (though again I don't understand how the error is handled, since it doesn't seem to be directly handled inside `cruisingModeStages.cpp`

Autopilot/PathManager/Src/cruisingMode/cruisingModeStages.cpp
Gongsta: got it

Autopilot/PathManager/Src/pathModeSelector.cpp
Gongsta: go it thanks!

Pop0097 (commented):
Cool! Looks like we can look into testing now. We can discuss writing tests for this sometime later this week once assignments stop killing me lol

Autopilot/PathManager/Inc/pathModeSelector.hpp
```diff
@@ -14,6 +14,12 @@
     enum PathModeEnum {MODE_TAKEOFF = 0, MODE_CRUISING, MODE_LANDING, MODE_TAXIING};
```
upadhyaydhruv, line 14: Later on we'll have to change this to account for the states more relevant to multicopters (hover, no taxiing, etc.). Do you think its easily changeable for that once its implemented?

Autopilot/PathManager/Src/cruisingMode/cruisingModeStages.cpp
```diff
@@ -17,14 +17,28 @@ _CruisingState_Telemetry_Return CruisingFlight::_return_to_ground;
 void CruisingFlight::execute(CruisingMode* cruise_mode) {
     Telemetry_PIGO_t telem_data = cruise_mode->getTelemetryData();
```
upadhyaydhruv, line 18: all this will now be from the jetson instead of the plane. When FW-CV Comms gets merged into devel, you will have to re-adapt this restructure

Autopilot/PathManager/Src/cruisingMode/cruisingModeStages.cpp
```diff
@@ -17,14 +17,28 @@ _CruisingState_Telemetry_Return CruisingFlight::_return_to_ground;
...
     SFOutput_t sf_data = cruise_mode->getSensorFusionData();
+    
 
     // Set waypoint manager input struct
     _input_data.track = sf_data.track; // Gets track
```
upadhyaydhruv, line 23: where is _input_data defined?

Autopilot/Test/Src/PathManager/Test_ModeSelector_Queue.cpp
```diff
@@ -0,0 +1,201 @@
+/*
```
upadhyaydhruv, line 1: Could you give a rundown of which tests you're running here (in the comments or in documentation) for future reference?

upadhyaydhruv (commented):
Great work! I added a few comments and only have a few concerns:
1.) Considering that a lot of PM will be restructured (implementation of FW-CV comms instead of ground telemetry, usage of other states like hover), do you think its best to hold off on integrating this, or can you easily add more elements/add changes later on?

Autopilot/PathManager/Inc/pathModeSelector.hpp
Gfcello: Think that makes more sense to sit in my PR for landing and takeoff

Autopilot/PathManager/Src/pathStateClasses.cpp
```diff
@@ -111,13 +111,14 @@ void modeExecutor::execute(pathManager* pathMgr)
             mode_selector->setCurrentMode(CruisingMode::getInstance());
         }
 
-        Telemetry_PIGO_t telem_data;
+        Telemetry_PIGO_t telem_data {};
 
+        // These were added to ensure CruisingMode_Fsm tests pass.
+        // TODO: Ensure CruisingMode tests don't depend on this so we can delete this garbage
```
Gfcello, line 117: Just commenting this TODO to keep it visible

Autopilot/Test/Src/PathManager/Test_TakeoffMode_Fsm.cpp
```diff
@@ -98,7 +99,10 @@ TEST(TakeoffModeFSM, TakeoffClimbTransitionsToCruising) {
     Telemetry_PIGO_t telem_data;
 	telem_data.holdingAltitude = 20;
 
-	SFOutput_t sf_data {};
+	SFOutput_t sf_data;
+	sf_data.altitude = 50; // So we transition to cruising mode
```
Gfcello, line 103: Once Hover is implemented, would this make more sense to transition to hover?

PR #158, opened 2021-12-25
Output mixing implementation

upadhyaydhruv:
Outputmixing has been implemented such that PWM values are set directly from the Attitude Manager. Various redundant states and modules have been removed, and all that remains is implementation of failsafe, and running the AM module in main.


upadhyaydhruv:
Accidentally made to devel instead of AM+C, closing and reopening correctly

PR #167, opened 2022-02-24
Dshot

Chrisc110:
# Description

Include a summary of the change and what was done. Also include relevant motivation and context for the change. If any, list dependencies are required for this change.

Include the Asana link, or relevant issue. (https://app.asana.com/0/187002107997407/list)

Don't forget to add reviewers on github and mention the PR in the appropriate discord channel.

## Testing

Please describe the tests that you ran to verify your changes. Provide instructions and details so the changes can be tested if required. Also mention who performed the testing.

## Documentation

# Merge Checklist:


Safety/Src/PWM.cpp
```diff
@@ -47,10 +53,109 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
+     *           b0 b1 b2 b3 b4 b5 b6 b7 b8 b9 b10 b11 b12 b13 b14 b15
+     *           |-------------------------------| |-| |-------------| 
+     *                      Throttle Data           ^     Checksum
+     *                                              |
+     *                                           Telemetry
+     */
+
+    uint16_t frame;
+
+    frame = ((DSHOT_MAX_THROTTLE * throttlePercentage) << 1) || (telemetry ? 1 : 0); //throttle and telemetry bits
```
Aidan-B: Throttle percentage isn't a decimal, it's an int. Your multiplication won't work here

Safety/Src/PWM.cpp
```diff
@@ -47,10 +53,109 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR2;
+        
+        case TIM_CHANNEL_3:
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR3;
+
+        case TIM_CHANNEL_4:
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR4;
+
+    }
+    HAL_DMA_Start_IT(dshotConfig.timer->hdma[dshotConfig.timDMAHandleIndex], (uint32_t)dshotConfig.dshotDMABuffer, destinationBuffer, DSHOT_DMA_BUFFER_SIZE);
```
Aidan-B: dshotDMABuffer is a pointer. I'm pretty sure you meant to dereference this, not cast it.

Safety/Inc/PWM.hpp
```diff
@@ -22,6 +22,26 @@
...
+	uint16_t num;
+	GPIO_TypeDef* port;
+	TIM_HandleTypeDef *timer;
+	uint16_t timer_channel;
+  bool isUsingDshot;
+  uint32_t *dshotDMABuffer;
+  uint16_t timDMAHandleIndex;
+  uint16_t timDMASources;
+
+} PWMPinConfig;
```
Aidan-B, line 39: Perhaps shift this below the PWMChannelNum and add some documentation comments

Safety/Inc/dshot.hpp
```diff
@@ -0,0 +1,75 @@
+/**
+ Implementation of the DSHOT protocol...
+ ect. add more detial on what dshot is here.
+ */
```
Aidan-B: Does this file ever ultimately get used? Should it be removed?

Safety/Src/PWM.cpp
```diff
@@ -36,6 +33,15 @@ PWMChannel::PWMChannel()
...
         PWMPinConfig currentChannel = PWM_CONFIG[i];
         HAL_TIM_PWM_Start(currentChannel.timer,currentChannel.timer_channel);
+
+        if (currentChannel.isUsingDshot)
+        {
+            //ANY DSHOT SETUP FUNCTIONS
+                //setting up timers?
+                //setting up dma call back function that runs when the transfer is complete and disables the dma
+                dshotSetupDMACallbacks();
+        }
```
Aidan-B, line 49: Don't need these comments any more?

Aidan-B (changes requested):

Safety/Src/PWM.cpp
Chrisc110: Right dshotDMABuffer _is_ a pointer and HAL_DMA_Start_IT takes a uint32_t address as that parameter so I just type casted a pointer to the buffer to a uint32_t. If I were to dereference the pointer I think I would be just giving it the value of the what ever is in index 0?

Safety/Inc/PWM.hpp
```diff
@@ -22,9 +22,29 @@
...
 
 #include <stdint.h>
+#include "stm32f0xx_hal.h"
+
+#define DSHOT_DATA_FRAME_LEN 16
+#define DSHOT_DMA_BUFFER_SIZE 18
+#define DSHOT_MAX_THROTTLE 2047
+#define DSHOT_BIT_1 240
+#define DSHOT_BIT_0 120
+#define NUM_DSHOT_MOTORS 4
```
sahil-kale: Why are these defined in the .hpp file as opposed to the .cpp file?

Also, we should be using variables as opposed to defines since this is C++ :)

Safety/Src/PWM.cpp
```diff
@@ -36,6 +33,13 @@ PWMChannel::PWMChannel()
     {
         PWMPinConfig currentChannel = PWM_CONFIG[i];
         HAL_TIM_PWM_Start(currentChannel.timer,currentChannel.timer_channel);
+
+        if (currentChannel.isUsingDshot)
+        {
+                //setting up dma call back function that runs when the transfer is complete and disables the dma
+                dshotSetupDMACallbacks();
+        }
```
sahil-kale: Wait hmmm shouldn't we be getting a pointer to what's in PWM_Config rather than an object? doesn't it create a copy of the object this way? I suppose it has the same effect in the end, but it would be better to do PWMPinConfig *currentChannel = &PWM_Config[i]

(This is not your fault LOL, in fact it might be mine, but I recommend editing it in this PR)

Safety/Src/PWM.cpp
```diff
@@ -47,10 +51,109 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
+
+void PWMChannel::dshotPrepareDMABuffer(uint32_t * dmaBuffer, uint8_t throttlePercentage)
+{
+    uint16_t frame = dshotPrepareFrame(throttlePercentage, false);
+
+    for (int i = 0; i < DSHOT_DATA_FRAME_LEN; i++)
+    {
+        dmaBuffer[i] = (frame & 0x8000 ? DSHOT_BIT_1 : DSHOT_BIT_0); //using the frame, populate the buffer with the duty cycle value corresponding to a 1 and 0
+
+        frame <<= 1;
```
sahil-kale: Ik what you're doing here, but a quick comment will help the youngins and others who read this understand better :)

Safety/Src/PWM.cpp
```diff
@@ -47,10 +51,109 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
+    }
+
+    dmaBuffer[16] = 0;
+    dmaBuffer[17] = 0;
```
sahil-kale, line 95: Purpose? Also, magicccc numbers

sahil-kale (changes requested):

Safety/Src/PWM.cpp
```diff
@@ -47,10 +51,109 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
+     *                                           Telemetry
+     */
+
+    uint16_t frame;
+
+    frame = ((uint16_t)(DSHOT_MAX_THROTTLE * (float)throttlePercentage / 100) << 1) || (telemetry ? 1 : 0); //throttle and telemetry bits
+
+    uint8_t checksum = (frame ^ (frame >> 4) ^ (frame >> 8)) & 0x00F; //calculating checksum... splitting first 12 bits into 3 nibbles and XORing
+
+    frame = (frame << 4) || checksum; //adding the checksum to the frame
```
Aidan-B: You want to use `|` (Bitwise OR) not `||` (Logical OR)
```suggestion
    frame = ((uint16_t)(DSHOT_MAX_THROTTLE * (float)throttlePercentage / 100) << 1) | (telemetry ? 1 : 0); //throttle and telemetry bits

    uint8_t checksum = (frame ^ (frame >> 4) ^ (frame >> 8)) & 0x00F; //calculating checksum... splitting first 12 bits into 3 nibbles and XORing

    frame = (frame << 4) | checksum; //adding the checksum to the frame
```

Safety/Src/PWM.cpp
```diff
@@ -47,10 +51,109 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR1;
+
+        case TIM_CHANNEL_2:
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR2;
+        
+        case TIM_CHANNEL_3:
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR3;
+
+        case TIM_CHANNEL_4:
+        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR4;
```
Aidan-B: You need to add a `break;` after each case, otherwise the `destinationBuffer` will always be `(uint32_t)&dshotConfig.timer->Instance->CCR4`

```suggestion
        case TIM_CHANNEL_1:
        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR1;
        break;

        case TIM_CHANNEL_2:
        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR2;
        break;

        case TIM_CHANNEL_3:
        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR3;
        break;

        case TIM_CHANNEL_4:
        destinationBuffer = (uint32_t)&dshotConfig.timer->Instance->CCR4;
        break;
```

Aidan-B (changes requested):
These changes are bugfixes that produce expected output on a scope (but not tested on a motor)

Safety/Src/PWM.cpp
Chrisc110: This leads to an error of "error: invalid conversion from 'const PWMPinConfig*' to 'PWMPinConfig*'"

Safety/Inc/PWM.hpp
Chrisc110: I have moved the #defines to the .cpp file, but is it convention to use constant variables rather than #define in cpp? I saw that MAX_CHANNELS and NUM_AVAILABLE_CHANNELS are #defines so I did the same?

Safety/Inc/PWM.hpp
```diff
@@ -44,6 +57,12 @@ typedef struct PWMGroupSetting {
 	bool inverted = false;
 } PWMGroupSetting;
 
+typedef enum dshotType
+{
+    DSHOT150,
+    DSHOT300,
+    DSHOT600
+} dshotType;
```
Aidan-B, line 65: It looks like this is never used. I assume you meant to use it somewhere during timing calculations?

Safety/Src/PWM.cpp
Aidan-B: It seems to me like you would just need to make `PWMPinConfig currentChannel` a const variable to fix this. eg. `const PWMPinConfig *currentChannel = &PWM_CONFIG[i];`

Safety/Src/PWM.cpp
```diff
@@ -3,31 +3,34 @@
 #include "PWM.hpp"
 #define MAX_CHANNELS 10
 #define NUM_AVAILABLE_CHANNELS 8
+#define DSHOT_DATA_FRAME_LEN 16
+#define DSHOT_DMA_BUFFER_SIZE 18
+#define DSHOT_MAX_THROTTLE 2047
+#define DSHOT_BIT_1 240
+#define DSHOT_BIT_0 120
```
Aidan-B: This isn't necessarily true, as it depends on the dshotType, right? or maybe that calculation is supposed to be done elsewhere? I would also appreciate the name being a little more descriptive of what it actually represents.

Safety/Src/PWM.cpp
```diff
@@ -47,10 +57,117 @@ void PWMChannel::set(uint8_t channel, uint8_t percent)
...
-    __HAL_TIM_SET_COMPARE((TIM_HandleTypeDef *) currentChannel.timer, currentChannel.timer_channel, (uint32_t) ticks);
 
+    if (currentChannel.isUsingDshot)
+    {
+        //prepare the buffers
+        dshotPrepareDMABuffer(currentChannel.dshotDMABuffer, percent);
+        //enable dma
+        dshotStartDMA(currentChannel);
+        //enable TIM dma requests?
+        dshotEnableDMARequests(currentChannel);
```
Aidan-B: Aren't these comments a bit redundant?

Safety/Src/PWM.cpp
```diff
@@ -3,31 +3,34 @@
 #include "PWM.hpp"
 #define MAX_CHANNELS 10
 #define NUM_AVAILABLE_CHANNELS 8
+#define DSHOT_DATA_FRAME_LEN 16
+#define DSHOT_DMA_BUFFER_SIZE 18
```
Aidan-B, line 7: Why does this need to be 18 instead of 16 (considering the frame length is 16). 
See also line 96 
```
//Setting the last two indexes in the buffer to 0 to ensure the last part of the signal is low
dmaBuffer[16] = 0;
dmaBuffer[17] = 0;
```

Safety/Inc/PWM.hpp
```diff
@@ -54,4 +73,57 @@ class PWMChannel {
...
 	uint32_t min_signal = 950; //standard for 50hz pwm signal, with a 100 Hz margin added to ensure we can reach the extreme values.
 	uint32_t max_signal = 2050;
+
+	/**
+  * @brief Prepares the Dshot data frame including the data, telemetry bit, and checksum
+  * @param throttlePercentage Trottle percentage output from the PID loop, 0-100%
+  * @param telemetry Bool indicating whether to set the telemetry bit or not
+  * @retval 16 bit dshot data frame
+  */    
+    uint16_t dshotPrepareFrame(uint8_t throttlePercentage, bool telemetry);
```
Aidan-B, line 83: should we be looking to change throttlePercentage to a float now that we have significantly more resolution with DSHOT? Maybe not for this PR, but something to consider

Safety/Inc/PWM.hpp
```diff
@@ -54,4 +73,57 @@ class PWMChannel {
...
+    void dshotSetupDMACallbacks();
 };
+
+
+/**
+  * @brief Callback function to be called when a DMA transfer is complete 
+  * @param hdma pointer to DMA handle
+  * @retval None
+  */
+    void dshotDMACompleteCallback(DMA_HandleTypeDef *hdma);
```
Aidan-B, line 128: I like the extensive documentation for each function here. Very nice 👍🏻

Aidan-B (changes requested):
Overall very well done. I'm excited to test this out on our new hardware. Most of my comments at this point are nitpicky, and hopefully are pretty easy to address.

Safety/Inc/PWM.hpp
Chrisc110: Yea right now it is not implemented anywhere to use this, but maybe as a future feature? I think at the clock speeds we are using, anything greater than 150 will be too fast...

Safety/Src/PWM.cpp
Chrisc110: This is because the first 16 bits is the actual dshot frame but we are reserving 2 extra bits to ensure the signal is LOW after the frame has passed. I'm actually not too sure if just 1 extra bit will do but that is what I have seen in other implementations

Safety/Inc/PWM.hpp
Chrisc110: That makes sense, I think with DSHOT we can definitely handle the increased sensitivity. I guess just a matter if it would actually make a difference or not...

Aidan-B:
I'm willing to mark this as approved as long as you add the things todo to asana:
- setting up the different dshot speeds
- changing pwm set from percentage to a more appropriate range (or a float)
- Anything else I missed

Just open it up as ready for review!

PR #176, opened 2022-04-17
Arm disarm

Chrisc110:
# Description

This PR is for the implementation of the Arm Disarm system.

## Testing

Please describe the tests that you ran to verify your changes. Provide instructions and details so the changes can be tested if required. Also mention who performed the testing.

## Documentation

# Merge Checklist:


Chrisc110:
Here is a rough drawing of where I have implemented this DisarmMode state class in the over all state machine: 
image

Chrisc110:
Here are a couple of questions I still have to complete this implementation:
1. When we are in the Disarm state, what pwm percentage should be sent to the motors? Should they be off? Should they be spinning at a very low rate? The values are currently set to 0.
2. Since Arm Disarm from the transceiver will just be a switch (either on or off) what range of PPM percentages should I accept as ARMED? Right now I have it as a percentage between 95-100% but I am not sure if this should be the case (due to my lack of knowledge of ppm in our system)
3. Since I have added a new PPM channel, was it correct for me to assume that the new channel in the PPM array will be the next index up? So right now "Channel 13" but index 12?

Safety/AttitudeManager/Inc/attitudeStateClasses.hpp
```diff
@@ -105,3 +106,20 @@ class FatalFailureMode : public attitudeState
...
+        void execute(attitudeManager* attitudeMgr);
+        void exit(attitudeManager* attitudeMgr) {(void) attitudeMgr;}
+        static attitudeState& getInstance();
+    private:
+        DisarmMode() {}
+        DisarmMode(const DisarmMode& other);
+        DisarmMode& operator =(const DisarmMode& other);
+        static bool ReceiveArmDisarmInstruction(attitudeManager* attitudeMgr);
+        static bool isArmed();
+        static uint8_t _armDisarmPPMValue;
```
upadhyaydhruv, line 123: Is this the reading on that specific PPM channel?

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -92,6 +101,21 @@ bool fetchInstructionsMode::ReceiveTeleopInstructions(attitudeManager* attitudeM
     return true;
 }
 
+bool fetchInstructionsMode:: isArmed()
+{
+    //Range of PPM percentage to be considered ARMED is 95-100%
+    const uint8_t MIN_ARM_VALUE = 50;
+    const uint8_t MAX_ARM_VALUE = 100;
+    if (_TeleopInstructions.PPMValues[ARM_DISARM_CHANNEL_INDEX] >= MIN_ARM_VALUE && _TeleopInstructions.PPMValues[ARM_DISARM_CHANNEL_INDEX] <= MAX_ARM_VALUE)
```
upadhyaydhruv: I would turn this into two lines for better readibility

upadhyaydhruv: Also, is checking that the value is less than the max arm value necessary? It shouldn't be necessary for a 2-state switch right?

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -223,3 +247,80 @@ attitudeState& FatalFailureMode::getInstance()
     return singleton;
```
upadhyaydhruv, line 245: Are you not checking if arm/disarm has been set here? You would have to check in every state right, in case the switch is flicked during any state?

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -223,3 +247,80 @@ attitudeState& FatalFailureMode::getInstance()
...
+{
+    static DisarmMode singleton;
+    return singleton;
+}
+
+bool DisarmMode:: ReceiveArmDisarmInstruction(attitudeManager* attitudeMgr)
+{
+    if(attitudeMgr->ppm->is_disconnected(HAL_GetTick()))
+    {
+        return false;
```
upadhyaydhruv: shouldn't have multiple returns in a function (MISRA C guidelines iirc)

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -223,3 +247,80 @@ attitudeState& FatalFailureMode::getInstance()
...
+}
+
+bool DisarmMode:: isArmed()
+{
+    //Range of PPM percentage to be considered ARMED is 95-100%
+    const uint8_t MIN_ARM_VALUE = 50;
+    const uint8_t MAX_ARM_VALUE = 100;
+    if (_armDisarmPPMValue >= MIN_ARM_VALUE && _armDisarmPPMValue <= MAX_ARM_VALUE)
+    {
+        return true;
```
upadhyaydhruv: Same here, multiple return statements undesirable (see https://pvs-studio.com/en/docs/warnings/v2506/)

upadhyaydhruv (commented):
Overall good work, just a few questions and comments on form/style, and some suggestions for improvement. Lmk if you have any questions!

Safety/AttitudeManager/Inc/attitudeStateClasses.hpp
Chrisc110: Yes.

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
Chrisc110: This comment has been resolved. Me and Dhruv agreed that it would be acceptable to only check in fetch instructions mode because the time delay would not inhibit any safety.

Safety/AttitudeManager/Inc/attitudeStateClasses.hpp
```diff
@@ -105,3 +106,20 @@ class FatalFailureMode : public attitudeState
...
+        static uint8_t armDisarmTimeoutCount;
+};
```
Aidan-B:
```suggestion
};

```

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -223,3 +246,79 @@ attitudeState& FatalFailureMode::getInstance()
...
+    //Range of PPM percentage to be considered ARMED is 95-100%
+    bool retVal = false;
+    const uint8_t MIN_ARM_VALUE = 50;
+    if (_armDisarmPPMValue >= MIN_ARM_VALUE)
+    {
+        retVal = true;
+    }
+
+    return retVal;
+}
```
Aidan-B, line 322:
```suggestion
}

```

Safety/Inc/PPM.hpp
```diff
@@ -21,6 +21,8 @@
 
 static const int32_t MAX_PPM_CHANNELS = 12;
 
+static const int32_t ARM_DISARM_CHANNEL_INDEX = 5 - 1; //5 - 1 becuase it is the fifth channel but the forth index in the array
```
Aidan-B: Arrays are usually 0 indexed. Do we need this math here?

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -223,3 +246,79 @@ attitudeState& FatalFailureMode::getInstance()
...
+    _armDisarmPPMValue = attitudeMgr->ppm->get(ARM_DISARM_CHANNEL_INDEX);
+
+    return retVal;
+}
+
+bool DisarmMode:: isArmed()
+{
+    //Range of PPM percentage to be considered ARMED is 95-100%
+    bool retVal = false;
+    const uint8_t MIN_ARM_VALUE = 50;
```
Aidan-B: You have an erroneous comment here, and you have re-defined this value from the other isArmed() function. I would define it once and reference it in both of these functions.

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -223,3 +246,79 @@ attitudeState& FatalFailureMode::getInstance()
...
+        3 possibilities:
+             1. Go into FatalFailureMode bec of timeout
+             2. Go into fetchInstructionsMode bec "Arm" instruction was sent
+             3. Do nothing, stay in the disarm state
+    */
+    if(armDisarmTimeoutCount > TIMEOUT_THRESHOLD && CommsFailed())
+    {
+        //Abort due to timeout failures
+        attitudeMgr->setState(FatalFailureMode::getInstance());
+        return;
```
Aidan-B, line 279: It seems to me that this code will basically boot the pilot into the fatalFailureMode if they are disarmed & disconnected. This seems rather odd to me? What if the pilot turns on the transmitter after they turn on the drone? What if it takes longer for the receiver to boot up then the safety chip takes to run these two iterations (not that I'm expecting that)? In my mind the disarm state would be the ONLY time that you would want to be able to disconnect the transmitter, all other times should lockout. Thoughts @upadhyaydhruv @antholuo?

Aidan-B (commented):
My biggest concern is how we handle the FatalFailureMode in the disarm state. Outside of that, I think this all looks very good and would be happy to try it out on hardware soon.

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
```diff
@@ -92,6 +101,20 @@ bool fetchInstructionsMode::ReceiveTeleopInstructions(attitudeManager* attitudeM
...
+    //Range of PPM percentage to be considered ARMED is 95-100%
+    const uint8_t MIN_ARM_VALUE = 50;
+    if (_TeleopInstructions.PPMValues[ARM_DISARM_CHANNEL_INDEX] >= MIN_ARM_VALUE)
+    {
+        return true;
+    }
+    else
+    {
+        return false;
+    }
```
Aidan-B: See Dhruv's previous comments
```suggestion
    bool retVal = false;
    if (_TeleopInstructions.PPMValues[ARM_DISARM_CHANNEL_INDEX] >= MIN_ARM_VALUE)
    {
        retVal = true;
    }
    return retVal;
```

Safety/AttitudeManager/Src/attitudeStateClasses.cpp
antholuo: If comms failed, would we be able to know if we have changed from arm to disarm?

Aidan-B: After talking with Anthony, I'm okay to leave this as is. Marking as resolved.

Aidan-B (approved):

PR #180, opened 2022-04-27
Fix RSSI

antimonyB:
# Description

RSSI was not working due to being implemented as a GPIO input rather than PWM. This PR implements the necessary PWM input according to https://www.team-blacksheep.com/tbs-crossfire-manual.pdf.

## Testing

Not tested.

## Documentation

# Merge Checklist:


Safety/Src/RSSI.cpp
```diff
@@ -4,22 +4,17 @@
 static volatile bool failed = false;
 static volatile int16_t counter = 0;
 
+//Time is in microseconds
+
+constexpr uint8_t MIN_ON_TIME = 1200; //20% of pwm signal from 1000 to 2000 microseconds
```
antholuo: Not sure if I'm ootl on this one, how did we decide 20%

Safety/Src/PPM.cpp
```diff
@@ -131,22 +135,57 @@ void HAL_TIM_IC_CaptureCallback(TIM_HandleTypeDef *htim)
...
+				index = 0;
+			}
+			else if (index < MAX_PPM_CHANNELS)
+			{
+				ppm_values[index] = pulseLength;
+				index++;
+			}
 		}
-		else if (index < MAX_PPM_CHANNELS)
+		else if (htim->Channel == HAL_TIM_ACTIVE_CHANNEL_2) //Recieve RSSI PWM
```
Aidan-B: This will never be called unless an appropriate `HAL_TIM_IC_Start_IT(&htim15, TIM_CHANNEL_2);` is also called during the init process.

Safety/Src/RSSI.cpp
Aidan-B: I'm also out of the loop. We should justify this number. I don't want to fall out of the sky if we still have a flyable connection, and likewise don't want to keep flying if the pilot cant really control anything. (I believe RSSI from our receiver is measured in dBm from 0 to -130 and so is not a linear scale)

Also note that 1200 is > than the max value for uint8_t, consider switching to a float to match the types for the comparison in UpdateRSSI

Aidan-B (dismissed):
There's also an issue in Arm/Disarm:
https://github.com/UWARG/ZeroPilot-SW/blob/23ebcbf5b120defa4f6928dcf117ce58719857eb/Safety/AttitudeManager/Src/attitudeStateClasses.cpp#L276
should become
`    if(armDisarmTimeoutCount >= TIMEOUT_THRESHOLD && CommsFailed())`

